PR #87, opened 2015-07-28
Better integration with FAF

mbrysa:
Retrace Server's repos are unfortunately still required for setting up the mock environment.


mtoman:
Thanks

PR #97, opened 2016-03-04
Replace yum with DNF

j-mracek:
The commit should provides replacement of yum by dnf. I am sorry but I didn't tested it by my self. I am not certain what import "from plugins import *" exactly do, therefore I only guess and then I try to change it according to the guess.

### DIFF
--- a/README
+++ b/README
@@ -14,17 +14,17 @@ itself consists of 2 parts:
   - guessparser:  Parser able to guess release from package's name. Can not
                   be relied on (e.g. el6 does not give enough information).
 
-  - yumcfg:       a string that will be appended to yum config file for all
+  - dnfcfg:       a string that will be appended to dnf config file for all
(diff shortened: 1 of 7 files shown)


jfilak:
Thank you for the patch. Unfortunately we cannot directly switch from yum to DNF because we must support distributions without DNF. We need to refactor the code and hide yum behind an abstraction and then add a new implementation using DNF.

j-mracek:
@jfilak Do you need some help with refactor from my side?

xsuchy:
@j-mracek Nope, we need to do that ourself.  Our server is on RHEL7 and we do not there DNF (well it is in EPEL, but...)

xsuchy:
This is on hold till RHEL8

trapas:

marusak:
Closed by  #218

PR #102, opened 2016-03-05
Rhbz1314973 refactor kernelver detection

DaveWysochanskiRH:
Addresses https://bugzilla.redhat.com/show_bug.cgi?id=1314973
With these patches, a fairly significant speedup is realized when users run retrace-server-interact <task> crash.

### DIFF
--- a/src/lib/retrace.py
+++ b/src/lib/retrace.py
@@ -713,100 +713,6 @@ def cache_files_from_debuginfo(debuginfo, basedir, files):
         cpio.wait()
         rpm2cpio.stdout.close()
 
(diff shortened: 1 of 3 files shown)


DaveWysochanskiRH:
Please don't merge this yet.  I would like to move a couple hunks from the second patch to the first patch, and make a third patch.  Having a patch that both moves code and changes code makes it difficult to see the change.  Also the small hunks in the second patch which removes the VMLINUX file should be in the first patch that introduces the vmlinux file.

DaveWysochanskiRH:
The new patches will be identical in content so if you are testing these two patches the testing will still be valid.  Thanks.

DaveWysochanskiRH:
After splitting, I found another small bug / omission in the second patch so will need to do some retest.

PR #105, opened 2016-03-10
Updating the stats handling and including a RHEL plugin

disneyworldguy:
Prior to the below commits, the release entries of the stats page were hard-coded to query and present the following releases.

   Fedora 15
   Fedora 16
   Fedora 17

The commits below alter this behaviour to allow the releases to be determined based off of the plugins present. For this process, The Fedora plugin has been refreshed to allow it to determine releases for all of the presently available Fedora releases. At the same time, a Red Hat Enterprise Linux plugin is also present to allow, though it currently does not include repository information to allow it to function with the retrace-server-reposync utility.


DaveWysochanskiRH:
I'm testing these two patches on top of the 3 patchsets from my pull requests and they look fine to me.

jfilak:
Thank you.

PR #107, opened 2016-03-11
Fix bt_filter missing last task/PID read if the last line was not blank

DaveWysochanskiRH:
This small patch to bt_filter was submitted by Steve Johnson and fixes the last task / pid missing in the bt_filter output


patriczek:
Looks good to me.

PR #108, opened 2016-03-14
manager.wsgi: Move the FTP query operation to an AJAX operation - Resolves rhbz1124462

disneyworldguy:
In the event that UseFTPTasks is enabled, and the backing FTP server
contains an exceptional number of files in it, the resulting query
can take an extreme amount of time to return. In the meantime, the
manager UI is unavailable for use to access existing tasks.

This patch implements an AJAX call to a new "ftp" interface. It will
return the remainder of the "FTP files" table by default, but will
also adhere to the "filter=" variable passing to limit the resulting
table.

Signed-off-by: Kyle Walker <walker.kyle.t@gmail.com>

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -32,6 +32,7 @@ interface_PYTHON = backtrace.wsgi \
                    index.wsgi \
                    log.wsgi \
                    manager.wsgi \
+                   ftp.wsgi \
                    settings.wsgi \
                    start.wsgi \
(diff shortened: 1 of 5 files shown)


DaveWysochanskiRH:
I tested this patch on top of some other patches and I confirm it solves the problem for the manager interface not being usable if there is a lot of files on the ftp server.

PR #110, opened 2016-03-15
Correct paths to Fedora development releases

jfilak:
All Fedora releases adopted Fedora Variants structure.

Signed-off-by: Jakub Filak <jfilak@redhat.com>


jfilak:
See https://bugzilla.redhat.com/show_bug.cgi?id=1317651

PR #117, opened 2016-05-26
Rhbz1340158 email notifications

DaveWysochanskiRH:
These minor fixup patches resolves issues with email notifications as described in https://bugzilla.redhat.com/show_bug.cgi?id=1340158

1. Typo causes email notifications to not get sent
2. After fixing typo, duplicate emails get sent for tasks that fail due to unable to determine kernel version

Tested these on a development server with these two patches added on top of retrace-server-1.15-1.el6.noarch


jfilak:
LGTM. Thank you!

PR #121, opened 2016-07-07
Add "exploitable" into LocationMatch in the httpd.conf

marusak:
Without this patch httpd refuses to download exploitable file.
On server the following message appears:
H01630: client denied by server configuration: /usr/share/retrace-server/exploitable.wsgi
And 403 is returned.
After creating this patch exploitable was returned correctly.


jfilak:
Thank you!

PR #124, opened 2016-07-20
Add python backtrace, source and locals into backtrace

marusak:
When python script fails, into backtrace file another information are
added. Specifically py-bt, py-locals and py-list.


jfilak:
Thank you!

PR #126, opened 2016-09-22
Fix coredump2packages

marusak:
While working with retrace-server I noticed, that sometimes coredump2packages returns less packages for very similar coredumps. Investigating this behaviour I found 2 problems.

1.  Packages with different architectures were marked as duplicity. Later one of them got deleted and often the one with architecture we wanted. Therefore it would be better to keep them both.

2. The eu-unstrip function returns different results based on ability to find ELF image. One of the cases is that FILE is set to '.' and MODULENAME is set to a file path. This case was ignored, even though that MODULNAME provides enough data.

### DIFF
--- a/src/coredump2packages
+++ b/src/coredump2packages
@@ -176,7 +176,7 @@ def process_unstrip_output():
         try:
             # if FILE (parts[2]) is not present on local filesystem
             # eu-unstrip uses FILE as MODULENAME (parts[4])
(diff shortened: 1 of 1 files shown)


marusak:
jfilak you were right about the loop being intentional. I just misunderstood what was happening there. Therefore I removed that one commit. Also I edited commit messages with more information. Hope that answers you question about FILE field.

jfilak:
Good job! Thank you.

Issue #115, opened 2016-04-01
Teach reposync to createa repo from FAF database


jfilak:
Finis the `UseFafPackages` efforts. We can use script like the following one which creates repo from RHEL-6 abrt builds:

```python
#!/usr/bin/python

import os
import sys
from argparse import ArgumentParser

import yum.misc

from createrepo import (MetaDataGenerator,
                        MetaDataConfig)

from pyfaf.storage import (getDatabase,
                           Build)

def get_pkglist(db, outputdir):
    q = db.session.query(Build).filter(Build.release.like("%.el6")).filter(Build.base_package_name == "abrt")
    for build in q.yield_per(1000):
        for pkg in build.packages:
            if pkg.has_lob("package"):
                print("Adding package: %s-%s-%s" % (pkg.name, build.version, build.release))
                yield os.path.relpath(pkg.get_lob_path("package"), outputdir)

def generate_yum_repo(db, outputdir):
    conf = MetaDataConfig()
    conf.quiet = False
    conf.verbose = True
    conf.update = False
    conf.sumtype = yum.misc._default_checksums[0]
    conf.outputdir = outputdir
    conf.directory = conf.outputdir
    conf.pkglist = get_pkglist(db, conf.directory)

    mdgen = MetaDataGenerator(config_obj=conf)

    if mdgen.checkTimeStamps():
        print("%s repo is up to date" % (reponame))

    mdgen.doPkgMetadata()
    mdgen.doRepoMetadata()
    mdgen.doFinalMove()
    mdgen.cleanup()

if __name__ == "__main__":
    parser = ArgumentParser(description="Generates yum repo from FAF databse")
    parser.add_argument("--outputdir", default=os.getcwd())
    args = parser.parse_args()

    print("Creating repo in '%s'" % (args.outputdir))

    generate_yum_repo(getDatabase(), args.outputdir)
```

PR #127, opened 2016-10-03
Enable creating repository from faf repository

marusak:
In faf there was introduced a way of assigning operating system with release
to builds. That makes it possible to find all builds of specific operating
system and release. Then repository can be created by using packages from faf
repository.

Action of creating repository from faf was integrated into
retrace-server-reposync, so when this action is executed and UseFafPackages
in configuration file is enabled, repository is created/updated from faf.

This closes abrt/retrace-server#115.

### DIFF
--- a/src/retrace-server-reposync-faf
+++ b/src/retrace-server-reposync-faf
@@ -1,0 +1,50 @@
+#!/usr/bin/python
+
+import os
+import sys
+from argparse import ArgumentParser
+import yum.misc
+from createrepo import MetaDataGenerator, MetaDataConfig
+from pyfaf.storage import getDatabase
+from pyfaf.queries import get_packages_by_osrelease
+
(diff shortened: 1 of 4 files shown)


src/retrace-server-reposync-faf
```diff
@@ -0,0 +1,49 @@
...
+import os
+import sys
+from argparse import ArgumentParser
+import yum.misc
+from createrepo import MetaDataGenerator, MetaDataConfig
+from pyfaf.storage import getDatabase
+from pyfaf.queries import get_packages_by_osrelease
+
+def get_pkglist(db, outputdir, opsys, release):
+    q = get_packages_by_osrelease(db, opsys, release)
```
jfilak: Don't we need to pass also the architecture?

jfilak:
Thank you!

Issue #77, opened 2015-06-19
Please use '-ascending' argument for gdb command 'thread apply all'


jfilak:
https://bugzilla.redhat.com/show_bug.cgi?id=1024504

PR #128, opened 2016-10-04
Add '-ascending' argument to gdb

marusak:
Use argument -ascending in gdbs "thread apply all" so the crashing thread is
depicted first.

Closes abrt/retrace-server#77.

### DIFF
--- a/src/plugins/fedora.py
+++ b/src/plugins/fedora.py
@@ -4,6 +4,8 @@ distribution = "fedora"
 abrtparser = re.compile("^Fedora release ([0-9]+) \(([^\)]+)\)$")
 guessparser = re.compile("\.fc([0-9]+)")
 displayrelease = "Fedora release"
+gdb_package = "gdb"
+gdb_executable = "/usr/bin/gdb"
 versionlist = [
   "fc1",
   "fc2",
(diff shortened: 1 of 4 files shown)


src/plugins/fedora.py
```diff
@@ -4,6 +4,9 @@
 abrtparser = re.compile("^Fedora release ([0-9]+) \(([^\)]+)\)$")
 guessparser = re.compile("\.fc([0-9]+)")
 displayrelease = "Fedora release"
+gdb = "gdb"
```
jfilak: I would call this member 'gdb_package' to avoid some confusion.

src/plugins/fedora.py
```diff
@@ -4,6 +4,9 @@
...
+gdb_run = "gdb -batch "
+gdb_stop = ""
```
jfilak: I would replace these two members with `gdb_executable` holding either `/usr/bin/gdb` or `/opt/rh/devtoolset-4/root/usr/bin/gdb`

marusak: This is great! Will change.

jfilak:
Thank you!

Issue #68, opened 2013-04-08
use a bit more user friendly message instead of "Killed by garbage collector"


jfilak:
Well, it is just for not so clever users like me. I didn't understand the message until it was explained me that the retrace job was running too long.

2013-04-07-15:18:43> Ok to upload core dump? (It may contain sensitive data). If your answer is 'No', a stack trace will be generated locally. (It may download a huge amount of data). 'YES'
2013-04-07-15:18:43  Querying server settings
2013-04-07-15:18:44  Preparing an archive to upload
2013-04-07-15:18:49  Uploading 5 megabytes
2013-04-07-15:18:57  Upload successful
2013-04-07-15:18:59  Retrace job started
2013-04-07-15:19:10  Analyzing crash data
2013-04-07-15:19:21  ................................................................................
2013-04-07-15:34:13  ................................................................................
2013-04-07-15:48:32  ...................................................................Retrace failed. Try again later and if the problem persists report this issue please.
2013-04-07-17:15:00  Killed by garbage collector

PR #129, opened 2016-10-11
Change error message in cleanup script

marusak:
Closes abrt/retrace-server#68

### DIFF
--- a/src/retrace-server-cleanup
+++ b/src/retrace-server-cleanup
@@ -93,7 +93,7 @@ if __name__ == "__main__":
                     continue
 
                 task.create_worker().clean_task()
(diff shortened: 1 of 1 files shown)


src/retrace-server-cleanup
```diff
@@ -93,7 +93,7 @@ if __name__ == "__main__":
                     continue
 
                 task.create_worker().clean_task()
-                task.set_log("Killed by garbage collector\n", True)
+                task.set_log("Task was killed due to running for too long or taking too many resources.\n", True)
```
jfilak: I'm not the best English speaker, but the for in "running for too long" sounds weird to me. Are you sure it should be there?

marusak: done

PR #136, opened 2016-12-05
Fixes on stats page

marusak:
Installing RS on new machine I discovered there were some bugs on stats page.
See commit messages to learn more about them.

### DIFF
--- a/src/stats.wsgi
+++ b/src/stats.wsgi
@@ -2,8 +2,23 @@
 from retrace import *
 sys.path.insert(0, "/usr/share/retrace-server/")
 
+status_queries = {"SELECT COUNT(*) FROM tasks": "{total}",
+    "SELECT COUNT(*) FROM tasks WHERE status = {0}".format(STATUS_SUCCESS): "{success}",
+    "SELECT COUNT(*) FROM tasks WHERE status = {0}".format(STATUS_FAIL): "{fail}",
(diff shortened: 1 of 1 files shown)


src/stats.wsgi
```diff
@@ -47,14 +47,13 @@ def application(environ, start_response):
...
     row = query.fetchone()
-    while row:
-        if int(row[0]) == STATUS_SUCCESS:
-            output = output.replace("{success}", str(row[1]))
-        elif int(row[0]) == STATUS_FAIL:
-            output = output.replace("{fail}", str(row[1]))
-        row = query.fetchone()
+    output = output.replace("{success}", str(row[0]))
+
+    query.execute("SELECT COUNT(*) FROM tasks WHERE status = {0}".format(STATUS_FAIL))
```
jfilak: How about to make the format string somehow shared? I am afraid that when we decide to update the table we will forget to update one of these formats.

marusak: I am not sure what you mean.  Until now all other items on this page are done the way I rewrote this -> execute some query, and with its response replace some tag in some string. These two I rewrote did it in loop, but when some value was not used, it was not replaced.

marusak: Ok, I've done some changes. Not you can just add another status into dictionary, and it will be changed correctly.

PR #138, opened 2016-12-12
Avoid circular dependency on kernel-debuginfo for vmlinux files alrea…

DaveWysochanskiRH:
…dy in cache

Today if the kernel-debuginfo repository goes away, any new vmcore queued will fail,
even if the vmlinux and other symbol files have already been extracted and stored locally.
The reason for this is there is a circular dependency inside prepare_debuginfo where we
unconditionally must find kernel-debuginfo for the kernel version in question to obtain
the location where the files should be stored.

Avoid this dependency by a simple check initially based on the "well known" path inside
most kernel-debuginfo files.  If the vmlinux file exists, just use this and avoid the
rest of prepare_debuginfo where we search for the kernel-debuginfo.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1368807

### DIFF
--- a/src/retrace/retrace.py
+++ b/src/retrace/retrace.py
@@ -1652,10 +1652,6 @@ class RetraceTask:
         if kernelver is None:
             raise Exception, "Unable to determine kernel version"
 
-        debuginfo = find_kernel_debuginfo(kernelver)
-        if not debuginfo:
-            raise Exception, "Unable to find debuginfo package"
-
         if "EL" in kernelver.release:
             if kernelver.flavour is None:
                 pattern = "EL/vmlinux"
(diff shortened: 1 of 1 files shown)


DaveWysochanskiRH:
If you're ok with this patch, is there any way to build this patch into a retrace-server-1.16 based release soon?  I would like to avoid the impact caused by the kernel-debuginfo NFS share going away at the end of December.  Also this should speed up tasks with existing extracted kernel-debuginfos rather than always having to lookup the kernel-debuginfo.  I tested the RHEL6 and RHEL7 kernels and with this patch retrace-server finds vmlinux in the cache and avoids searching for the kernel-debuginfos.  The rest of the code remains in tact so in theory there should be no regressions caused by this.  I tested by removing the extracted path for one kernel and re-submitting it and the kernel-debuginfo was found and extracted as before.

jfilak:
The patch looks good to me. Good job. I will try to release new version ASAP.

PR #139, opened 2017-01-21
Improve autogen to list and install dependencies

marusak:
Signed-off-by: Matej Marusak <marusak.matej@gmail.com>

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -1,27 +1,72 @@
 #! /bin/sh
 
-mkdir -p m4
-echo "Creating m4/aclocal.m4 ..."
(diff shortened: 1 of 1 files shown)


jfilak:
Good job. Other ABRT projects run ``./configure`` at the end of ``./autogen.sh``, what about to do the same here too? Please notice that the other autogen.sh scripts configures sources for local debugging by default. It would be nice if you could follow this rule.

marusak:
@jfilak Thanks for review. I edited it as you suggested. Is it now ok?

jfilak:
LGTM. Thank you!

PR #140, opened 2017-01-26
Update readme

marusak:
Update Readme to contain information about project what it is, what it does and how it works. Also add Contributing page (it is from abrt with slight changes) and Deploying page.

Will close #135 

And also in this PR is one other commit. It does not have anything to do with PR, but I noticed it and kept it in the same branch. If there is problem with it, I can split it into two PR, or push the other commit directly.


jfilak:
Cool as an iceberg!

PR #142, opened 2017-02-15
Default to hex mode for crash commands involving backtraces.

DaveWysochanskiRH:
The kernel log contains an oops message which is printed with hex offsets.
For consistency, use hex when calling crash commands involving backtraces.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1416086

### DIFF
--- a/src/retrace/retrace_worker.py
+++ b/src/retrace/retrace_worker.py
@@ -652,7 +652,7 @@ class RetraceWorker(object):
                         log_warn("crash 'log' exitted with %d" % child.returncode)
 
                     child = Popen(["/usr/bin/mock", "--configdir", cfgdir, "shell", "--",
-                                   "crash -s %s %s" % (vmcore, vmlinux)], stdin=PIPE, stdout=PIPE, stderr=null)
+                                   "crash --hex -s %s %s" % (vmcore, vmlinux)], stdin=PIPE, stdout=PIPE, stderr=null)
                     crash_bt_a = child.communicate("bt -a\nquit\n")[0]
                     if child.wait():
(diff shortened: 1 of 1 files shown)


jfilak:
Dave, what version of crash do I need to have installed? `crash-7.1.5-2.el7.x86_64` does not support `--hex` parameter, or am I doing something wrong?

marusak:
@jfilak I have installed the same version as you and I have `--hex` available. Does crash return `crash: invalid option: --hex`?
@DaveWysochanskiRH This PR looks good to me. However I am not skilled in field of retracing vmcores. But if you believe that having hexadecimal is better, I am happy to merge this. Only one question comes to my mind (when jfilak wrote that his version does not support --hex) - Was this parameter recently added? And therefore can cause problems when older crash will be available?

DaveWysochanskiRH:
@jfilak crash changelog is here http://people.redhat.com/anderson/crash.changelog.html and shows --hex being added on 6.0.9 so the version you have should definitely work.  What error were you getting?
@marusak Yeah to me this is a minor issue, but we have had multiple people complain about the usage of decimal in retrace-server output and it usually involves the backtraces and offsets of functions (for example, the RIP symbol).  Marc who filed bugzilla 1416086 is one of the senior engineers handling vmcores on a daily basis.

DaveWysochanskiRH:
@jfilak I guess it somewhat depends on how far back in compatibility we need to go.  For example, early RHEL6 builds crash version was crash-5.x so this would not work.  Later RHEL6 builds are crash-6.1.x so it should have --hex.  Let me do another patch with the other option of using "set hex" inside crash so you can discard this patch for now.

PR #143, opened 2017-02-16
Include md5sum of original archive in summary page

marusak:
This commit adds option for users to see md5sum of original archive. User can
select before submitting task whether this information will be displayed or not.

If selected, this information is displayed on summary page as well as is written
into file. Format of this file is same as when md5sum command is run, so these
data can be compared easily (for example with diff).

Also into configuration file new item was added, to have this checkbox checked
by default.

### DIFF
--- a/src/manager_vmcore_task_form.xhtml
+++ b/src/manager_vmcore_task_form.xhtml
@@ -10,4 +10,5 @@
         <input type="submit" value="Create vmcore task" class="submit" />
         <div class="center">Any URL that <code>wget</code> can download or a local path (<code>file:///foo/bar</code> or just <code>/foo/bar</code>)</div>
       </div>
+      <input type="checkbox" {md5_enabled} name="md5sum" class="checkbox" /><span class="formsubhead"> Calculate md5 checksum for all downloaded resources</span>
     </form>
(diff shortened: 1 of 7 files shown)


DaveWysochanskiRH:
I tested this on our dev server and ran through the following 12 different cases, and they all worked very well.  I think this should be merged.
Test case variables
- CalculateMd5: 0/1
- webUI: 'local file' / 'ftp file'
- commandline tool (queue-in-retrace.py): 'local file'

src/manager_vmcore_task_form.xhtml
```diff
@@ -10,4 +10,5 @@
         <input type="submit" value="Create vmcore task" class="submit" />
         <div class="center">Any URL that <code>wget</code> can download or a local path (<code>file:///foo/bar</code> or just <code>/foo/bar</code>)</div>
       </div>
+      <input type="checkbox" {md5_enabled} name="md5sum" class="checkbox" /><span class="formsubhead"> Calculate md5 checksum for all downloaded resources</span>
```
xsuchy: It will be better if there will be "label" tag instead of "span"
   https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_label
But it that would require change everywhere else in form too. So for this PR this should be ok.

PR #144, opened 2017-02-16
Default to hex mode for crash commands involving backtraces.

DaveWysochanskiRH:
The kernel log contains an oops message which is printed with hex offsets.
For consistency, output in hex when running crash commands involving backtraces.
Use the "set hex" command inside crash to ensure the best backward compatibility
with crash versions.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1416086


DaveWysochanskiRH:
Tested patch on our test/dev retrace-server instance which is running crash-7.1.4-1.el6_6.x86_64 and retrace-server-1.16-3.el6.noarch (plus this patch)

PR #145, opened 2017-02-22
Modify search for existing vmlinux files in cache to handle older ker…

DaveWysochanskiRH:
…nel-debuginfos

The initial patch handles the latest kernel-debuginfos since these are predictable.
However, older kernel-debuginfos (rhel5, etc) have a few different variations of paths
to vmlinux files, so make sure we handle these.  In addition, handle special cases
such as debug kernels and other variants.

### DIFF
--- a/src/retrace/retrace.py
+++ b/src/retrace/retrace.py
@@ -1665,16 +1665,33 @@ class RetraceTask:
             os.makedirs(debugdir_base)
         #
         # First look in our cache at the "typical" location which is something like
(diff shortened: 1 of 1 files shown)


DaveWysochanskiRH:
I tested this patch very thoroughly on 33 different vmcores and it should handle all cases of kernel-debuginfos.  The original patch handled mainly newer kernels (rhel6, rhel7) but did not handle rhel5 or earlier, nor did it handle special variants such as debug kernels.  I also tested by removing the existing cache and resubmitting all vmcores to ensure the existing logic of rpm2cpio extraction was not affected by these caching patches.

PR #149, opened 2017-03-31
Fix problem with missing modules on kernel versions with cached vmlin…

DaveWysochanskiRH:
…ux files.

The first few caching patches just looked for the vmlinux file in the cache, and
if the vmlinux file existed, it would return early from prepare_debuginfo, skipping
the section of finding the debuginfo, listing the modules, and extracting the modules
if required.  However, this approach is naive and wrong.  Even if the vmlinux file
exists in the cache, this does not mean all module files also exist in the cache.
The module extraction code only extracts modules in use for a given vmcore.

To avoid the module problem, for now just unconditionally look for the kernel-debuginfo
as before, but don't fail the task if the cached vmlinux exists but for some reason
the debuginfo does not.  This was the original issue that prompted Red Hat bug 1368807.

A future optimization could be to check the cache for all required module files, and
if they all exist, skip find of the kernel-debuginfo.  For now just keep the optimization
limited to the vmlinux file.  Note that in the case where all module files exist in
the cache, cache_files_from_debuginfo should return early without doing anything.
So the overhead with the modules in the cache would only be the finding of the
kernel-debuginfo and listing the files in the debuginfo.  My guess is in a lot of
vmcores with known kernels and cached vmlinux files will have a different module
footprint, so it's unclear whether additional refactoring would really help much.

### DIFF
--- a/src/retrace/retrace.py
+++ b/src/retrace/retrace.py
@@ -1665,21 +1665,12 @@ class RetraceTask:
         if kernelver is None:
             raise Exception, "Unable to determine kernel version"
 
-        if "EL" in kernelver.release:
-            if kernelver.flavour is None:
-                pattern = "EL/vmlinux"
-            else:
-                pattern = "EL%s/vmlinux" % kernelver.flavour
-        else:
-            pattern = "/vmlinux"
-
         debugdir_base = os.path.join(CONFIG["RepoDir"], "kernel", kernelver.arch)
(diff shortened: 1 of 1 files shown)


DaveWysochanskiRH:
This fixes the problem as described in https://bugzilla.redhat.com/show_bug.cgi?id=1368807#c14

xsuchy:
👍

PR #154, opened 2017-05-10
Add commandline client

marusak:
This commit implements commandline client for creating, getting and
setting tasks.

### DIFF
--- a/retrace-server.spec.in
+++ b/retrace-server.spec.in
@@ -38,6 +38,8 @@ Requires: crash >= 5.1.7
 Requires: wget
 Requires: kexec-tools
 Requires: python-magic
+Requires: python-requests
+Requires: python-requests-kerberos
 Requires(preun): /sbin/install-info
 Requires(post): /sbin/install-info
 Requires(post): /usr/bin/crontab
@@ -125,6 +127,7 @@ fi
 %{_bindir}/%{name}-reposync
 %{_bindir}/%{name}-reposync-faf
 %{_bindir}/%{name}-plugin-checker
+%{_bindir}/%{name}-task
 %{_bindir}/bt_filter
 %{_bindir}/coredump2packages
 %{python_site}/retrace/*
@@ -133,6 +136,7 @@ fi
 %doc %{_mandir}/man1/%{name}-interact.1*
 %doc %{_mandir}/man1/%{name}-reposync.1*
 %doc %{_mandir}/man1/%{name}-worker.1*
+%doc %{_mandir}/man1/%{name}-task.1*
 %doc %{_infodir}/%{name}*
 %doc COPYING README.md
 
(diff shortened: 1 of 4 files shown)


marusak:
@mkutlak Could you please test this and let me know if you find any problems? Thanks.

xsuchy:
Can you write it python3 friendly (so future migration is easier)? I.e. do not use xrange. Do not use bare except. And there are some style issue (pylint-3 is your friend).

jfilak:
I thought the retrace-server client is already implemented:
https://github.com/abrt/abrt/blob/master/src/plugins/abrt-retrace-client.c

marusak:
> I thought the retrace-server client is already implemented:
https://github.com/abrt/abrt/blob/master/src/plugins/abrt-retrace-client.c

Yeah, when I first got into this, I also was proposing that retrace-client exists. But this code does much more.
Few of them:
- Support of manager and ftp tasks (the main request from @DaveWysochanskiRH)
- Support for kerberos
- Support for interactive tasks (great for debugging)
- Does not need valid dumpdir (the abrt's script requires some other files such as 'type' and 'time'
- Get and Set options for different fields (status, email, caseno...)

jfilak:
Alright then. So I propose to create a separate package for the client and drop the ABRT's client. There is no need to maintain two clients, or do I miss something?

jfilak:
Maybe it is not such a good idea. ABRT client does some tasks specific only to ABRT - for example checking the correctness of the directory contents based on the problem type.

DaveWysochanskiRH:
Hi yeah the original bug is https://bugzilla.redhat.com/show_bug.cgi?id=1396608 (I just added jakub@thefilaks.net on CC list so you can see it Jakub)
The main needs are:
1. to submit a local vmcore (kernel core) to the local retrace-server
2. handle kerberos (this was a requirement that came shortly after the bug was filed due to added security requirements)

DaveWysochanskiRH:
I think the use case is slightly different between this and abrt-retrace-client, but agree they are similar.

marusak:
> Alright then. So I propose to create a separate package

I agree with this part. It would be nice to have all cli actions from RS under one package. Or maybe even better create one RS cli and make actions pluginable as it is in FaF. But that is for some later.

>  drop the ABRT's client. There is no need to maintain two clients, or do I miss something?
Maybe it is not such a good idea. ABRT client does some tasks specific only to ABRT - for example checking the correctness of the directory contents based on the problem type.

Yeah, it is not such a good idea. We probably could merge into one, but I think the way there are today is fine. ABRT's one does things that are required from ABRT's perspective, and this does thinks more general for RS.

marusak:
@xsuchy wrote:

> Can you write it python3 friendly (so future migration is easier)? I.e. do not use xrange. Do not use bare except. And there are some style issue (pylint-3 is your friend).

Sure I can. The only non-python3 thing was `xrange`. Changed it.
This is weird, I have pylint both on my work computer and my personal computer. It seemed OK on one and second one displayed quite a few problems. Fixed them (except 3 minor those I dont care about).

PR #158, opened 2017-05-18
Update mock configuration

mkutlak:
Since mock-1.4.1, BindMount plugin is required to make directories available
for baseurl=file://.

https://github.com/rpm-software-management/mock/wiki#using-file-urls-in-configs

And use DNF in mock configuration on Fedora.


xsuchy:
Please do not call platfom.linux_distribution.
See http://miroslav.suchy.cz/blog/archives/2016/10/31/how_to_replace_deprecated_platform_distro/index.html

xsuchy:
LGTM

PR #164, opened 2017-11-27
Set kernelver and vmlinux as soon as possible

DaveWysochanskiRH:
For larger vmcores it is possible that prepare_debuginfo may take a long time.
This is due to various factors such as makedumpfile, as well as how long crash
takes to load when detecting modules as well as running the early commands
such as 'sys'.  Today we are not setting kernelver or vmlinux as early as we
can and as a result, someone waiting to get access to the vmcore either has to
wait excessivly long or do some manual intervention like setting up vmlinux
manually.  Today if someone tries to run 'retrace-server-interact' before all
commands complete in the retrace, they will end up blocking waiting
for the re-detction of the kernelver which can take a very long time (since
it may run strings on the whole file).  This makes no sense since retrace
figures out the kernelver early on as well as the vmlinux file (assuming it
is in the cache).  This patch sets both the kernelver file and the vmlinux
file as early as reasonably possible, and thus allows much earlier access
to the vmcore through 'retrace-server-interact <task> crash'.

Resolves https://bugzilla.redhat.com/show_bug.cgi?id=1516329

### DIFF
--- a/src/retrace/retrace.py
+++ b/src/retrace/retrace.py
@@ -1670,6 +1670,8 @@ class RetraceTask:
         if kernelver is None:
             raise Exception, "Unable to determine kernel version"
 
+        self.set_kernelver(str(kernelver))
+
         debugdir_base = os.path.join(CONFIG["RepoDir"], "kernel", kernelver.arch)
         if not os.path.isdir(debugdir_base):
             os.makedirs(debugdir_base)
@@ -1696,6 +1698,7 @@ class RetraceTask:
         if os.path.isfile(vmlinux_cache_path):
             log_info("Found cached vmlinux at path: " + vmlinux_cache_path)
             vmlinux = vmlinux_cache_path
+            self.set_vmlinux(vmlinux)
         else:
             log_info("Unable to find cached vmlinux at path: " + vmlinux_cache_path)
             vmlinux = None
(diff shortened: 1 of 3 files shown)


src/retrace/retrace_worker.py
```diff
@@ -561,9 +561,8 @@ def start_vmcore(self, custom_kernelver=None):
 
             log_debug("Determined kernel version: %s" % kernelver)
 
-        kernelver_str = kernelver.kernelver_str
-
         task.set_kernelver(str(kernelver))
+        kernelver_str = kernelver.kernelver_str
```
marusak: Why the changed order?

src/retrace/retrace.py
```diff
@@ -1754,16 +1756,13 @@ def prepare_debuginfo(self, vmcore, chroot=None, kernelver=None, crash_cmd=["cra
         # Note the dependency from this code on the debuginfo file list
         if vmlinux is None:
             vmlinux_debuginfo = os.path.join(debugdir_base, vmlinux_path.lstrip("/"))
+            cache_files_from_debuginfo(debuginfo, debugdir_base, [vmlinux_path])
```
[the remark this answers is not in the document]
DaveWysochanskiRH: The call is prefaced by 'if vmlinux is None:'.  So if that is true, we don't have a vmlinux file yet in the cache (see prior code that looks in the standard location for it).  Then even if cache_files_from_debuginfo is called, we still need to check if the file exists (because it's possible cache_files_from_vmlinux may silently fail).  In short, yes is it necessary and no it is not unconditionally called.  I hope that explains it.  If you're thinking of something else and I missed what you're saying please clarify maybe with further code suggestion.  Thanks.

src/retrace/retrace_worker.py
DaveWysochanskiRH: It is a minor change which moves the setting of kernelver a bit earlier and goes with the theme of the patch "Set kernelver and vmlinux as soon as possible"

PR #169, opened 2018-01-17
Remove duplicate RetraceWorker._fail call when start_vmcore fails

DaveWysochanskiRH:
This patch fixes https://bugzilla.redhat.com/show_bug.cgi?id=1535485 
Calling _fail inside start_vmcore() is redundant since the only caller, start() also calls _fail() inside an exception handler.so just raise an exception and allow start() to call fail.

### DIFF
--- a/src/retrace/retrace_worker.py
+++ b/src/retrace/retrace_worker.py
@@ -641,16 +641,14 @@ class RetraceWorker(object):
                 os.symlink("/etc/mock/logging.ini",
                            os.path.join(task.get_savedir(), RetraceTask.MOCK_LOGGING_INI))
             except Exception as ex:
-                log_error("Unable to create mock config file: %s" % ex)
-                self._fail()
+                raise Exception("Unable to create mock config file: %s" % ex)
             finally:
                 os.umask(old_umask)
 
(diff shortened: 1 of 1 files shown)


src/retrace/retrace_worker.py
```diff
@@ -725,8 +723,7 @@ def start_vmcore(self, custom_kernelver=None):
                         crash_foreach_bt = None
 
             except Exception as ex:
-                log_error(str(ex))
-                self._fail()
+                raise Exception(str(ex))
```
marusak: Does this make sense? To catch it and then re-raise it immediately?

DaveWysochanskiRH: As far as I understand this is the way you allow a caller to handle this exception which is what we want here, though I'm no Python expert.  Do you see an alternative or see a problem?

DaveWysochanskiRH: Ah yes ok I see the point.  I think the right thing to do there is to have a log message and raise the exception there similar to the others.  I will resubmit.

DaveWysochanskiRH:
Will resubmit with minor changes

PR #170, opened 2018-01-19
Remove duplicate RetraceWorker._fail call when start_vmcore fails

DaveWysochanskiRH:
The only place start_vmcore is called from is the start method of
RetraceWorker, and that call is inside a try ... except clause which
calls _fail in the 'except' part.  Thus, inside start_vmcore, if we
have a nested exception, we don't need to call _fail there.  Instead
we should just raise an exception and let the 'start' method call
_fail with our Exception string set to the describe the failure.

This patch also adds an error message if mock is being used and
prepare_debuginfo or one of the crash commands fail.  This makes
it consistent with the non-mock case.  Once we move this except
clause un-indent the rest of the block to make it consistent.

### DIFF
--- a/src/retrace/retrace_worker.py
+++ b/src/retrace/retrace_worker.py
@@ -662,16 +662,14 @@ class RetraceWorker(object):
                 os.symlink("/etc/mock/logging.ini",
                            os.path.join(task.get_savedir(), RetraceTask.MOCK_LOGGING_INI))
             except Exception as ex:
-                log_error("Unable to create mock config file: %s" % ex)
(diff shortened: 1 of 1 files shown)


DaveWysochanskiRH:
This is a fixed up version of the original patch.  I fixed the one 'mock' prepare_debuginfo failure exception to be more consistent with the 'non-mock' scenario which is why the patch is larger now (I had to un-indent the whole block of the mock commands).  Now both "try .. except" clauses only catch the failure in prepare_debuginfo - originally they were inconsistent.

PR #172, opened 2018-02-01
update spec file

xsuchy:


DaveWysochanskiRH:
FYI - due to various delays from system upgrades beyond my control, we are still running RHEL6 so I'm having to manually take out patches like this.

PR #174, opened 2018-02-02
Rhbz1535592 rhbz1440381 kernelver parser bz1540253 crash osrelease spin final

DaveWysochanskiRH:
These two patches fix a few bugs in the get_kernel_release() function which detects the kernel version from a vmcore.  It is a critical function so I went through a number of iterations and performed significant testing.

The testing showed there were no regressions in kernel detection on a few thousand production vmcores.  In addition I have these two patches running now on our development retrace-server which also has a auto-submit script and it is currently mirroring the vmcores that come in to production.  So far I've seen no regressions there in the new vmcores vs production.

I request review and pull of these patches as soon as reasonable.  I would like to schedule an update to the production retrace-server with these patches as well as most all of the latest upstream code.

Thanks.

### DIFF
--- a/src/retrace/retrace.py
+++ b/src/retrace/retrace.py
@@ -89,12 +89,6 @@ REPODIR_NAME_PARSER = re.compile("^[^\-]+\-[^\-]+\-[^\-]+$")
 
 KO_DEBUG_PARSER = re.compile("^.*/([a-zA-Z0-9_\-]+)\.ko\.debug$")
 
-# parsers for vmcore version
-# 2.6.32-209.el6.x86_64 | 2.6.18-197.el5
-KERNEL_RELEASE_PARSER = re.compile("^([0-9]+\.[0-9]+\.[0-9]+)-([0-9]+\.[^ \t]*)$")
-# OSRELEASE=2.6.32-209.el6.x86_64
-OSRELEASE_VAR_PARSER = re.compile("^OSRELEASE=([^%]*)$")
-
 DUMP_LEVEL_PARSER = re.compile("^[ \t]*dump_level[ \t]*:[ \t]*([0-9]+).*$")
 
 WORKER_RUNNING_PARSER = re.compile("^[ \t]*([0-9]+)[ \t]+[0-9]+[ \t]+([^ ^\t]+)[ \t]+.*retrace-server-worker ([0-9]+)( .*)?$")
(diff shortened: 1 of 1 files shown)


src/retrace/retrace.py
```diff
@@ -477,66 +472,92 @@ def is_package_known(package_nvr, arch, releaseid=None):
...
+# in the first portion of the file.  Testing has indicated that we do
+# not need to scan the entire file but can rely on a small portion
+# at the start of the file, which helps preserve useful pages in the
+# OS page cache.
+#
+# The following regex's are used for the 40% scenario
+# 1. Look for 'OSRELEASE='.  For example:
+# OSRELEASE=2.6.18-406.el5
+# NOTE: We can get "OSRELEASE=%" so we disallow the '%' character after the '='
+OSRELEASE_VAR_PARSER = re.compile("OSRELEASE=([^%][^\x00\s\n\r]+)")
```
marusak: [^\x00\s\n\r] means - "does not contain NULL, whitespace-characters, new line and return". However \s is equivalent to [ \t\n\r\f\v] so the "\n\r" is redundant. Also note that there is \S which is [^ \s]. Only reason why not using \S is the \x00. Why is there this NULL character? I cannot quite imagine where this could pop up.

src/retrace/retrace.py
```diff
@@ -477,66 +472,92 @@ def is_package_known(package_nvr, arch, releaseid=None):
...
+# 2. Look for "Linux version" string.  For exmaple:
+# Linux version 3.10.0-693.11.1.el7.x86_64 (mockbuild@x86-041.build.eng.bos.redhat.com) (gcc version 4.8.5 20150623 (Red Hat 4.8.5-16) (GCC) ) #1 SMP Fri Oct 27 05:39:05 EDT 2017
+LINUX_VERSION_PARSER = re.compile('Linux\sversion\s(\S+)\s+(.*20\d{1,2}|#1\s.*20\d{1,2})')
```
marusak: This matches in first group `3.10.0-693.11.1.el7.x86_64` and in second group `(mockbuild@x86-041.build.eng.bos.redhat.com) (gcc version 4.8.5 20150623 (Red Hat 4.8.5-16) (GCC) ) #1 SMP Fri Oct 27 05:39:05 EDT 2017`, right? Is the second group needed? It seems to me, that it is rather useless, but please enlighten me.

src/retrace/retrace.py
```diff
@@ -477,66 +472,92 @@ def is_package_known(package_nvr, arch, releaseid=None):
...
+# 3. Look for the actual kernel release. For example:
+# 2.6.32-209.el6.x86_64 | 2.6.18-197.el5
+KERNEL_RELEASE_PARSER = re.compile('([0-9]+\.[0-9]+\.[0-9]+)-([0-9]+\.[^\x00\s\n\r]+)')
```
marusak: Wouldn't it be better readable to use \d instead of [0-9]? Also same about [^\x00\s\n\r] as above.

src/retrace/retrace.py
```diff
@@ -477,66 +472,92 @@ def is_package_known(package_nvr, arch, releaseid=None):
...
+#
```
marusak: line with empty comment, I guess that it is only forgotten

src/retrace/retrace.py
DaveWysochanskiRH: Null is needed as a null terminator of the string.  I originally did not have null but had to add it to ensure the regex used with the open/read method was equivalent to the regex used with the 'strings' method.  Remember the goal is to remove 'strings' and so I'm just trying to not lose any detection while improving it.  I actually was skeptical myself until I saw one vmcore and opened it to find that the kernel version was terminated by a NULL (I think it was a RHEL5 version).  If you think this is not clear, I may be able to break this into a "matches" regex and a "non-matches" regex:
 +OSRELEASE_VAR_PARSER = re.compile("OSRELEASE=([^%][\S]+)[\x00\s]+")

I have not tested the above and so it would need another round of testing,  but I think that would work.  Is that clearer? If you think it's clearer I am ok with doing another round of testing.

src/retrace/retrace.py
DaveWysochanskiRH: This whole regex was taken from the CAS code as-is to detect one definitive signature of a vmcore which is that "Linux version" line. I am not sure what removal of the second group would do but at the very least it would weaken the regex.  How significant this is - I cannot really say.  I'd prefer to leave it as is.

src/retrace/retrace.py
DaveWysochanskiRH: I don't think it matters but I can easily change this.

src/retrace/retrace.py
marusak: > until I saw one vmcore and opened it to find that the kernel version was terminated by a NULL

Ok, when it is required that it is fine to have

When we need to have the `\x00` then I would not go into `\S` only would drop `\n\r`, so the result would be
`OSRELEASE_VAR_PARSER = re.compile("OSRELEASE=([^%][^\x00\s]+)")`

src/retrace/retrace.py
marusak: > This whole regex was taken from the CAS 

Ok, we can leave as it is, I quess that they have good reason to write it that way. Could you please write that into comment that where it was taken from? Ideally with github or where they keep their source codes, so we know where to look, when something goes wrong.

src/retrace/retrace.py
marusak: `KERNEL_RELEASE_PARSER = re.compile('(\d+\.\d+\.\d+)-(\d+\.[^\x00\s]+)')`

marusak:
Can you please update those small changes mentioned above? Then it does not need anymore testing and we can merge it. @DaveWysochanskiRH 

Thank you

src/retrace/retrace.py
DaveWysochanskiRH: ok yeah that sounds good

src/retrace/retrace.py
DaveWysochanskiRH: Ok sure I can add a comment in the code.  FYI, this is also given in https://bugzilla.redhat.com/show_bug.cgi?id=1535592#c9

src/retrace/retrace.py
DaveWysochanskiRH: ok sure

DaveWysochanskiRH:
I have fixed these up and will close this pull request and resubmit via different branch.  Thanks!

PR #178, opened 2018-03-06
Fail task if the crash sys command exits with non-zero and kernellog …

DaveWysochanskiRH:
…is small

Historically we have only checked to see if we can detect the kernel version
for a vmcore and if we could then we declare a task as 'success'.  However,
there are a number of instances where we cannot even obtain a kernel log
file and so it is very likely the vmcore is useless.  In these cases it makes
more sense to just declare the task as failed.

Check for a non-success exit code when running the 'sys' command of crash.
If 'sys' fails, and we have a very small kernel log, it's likely the vmcore
is useless so fail the task.

Once the above is added, we have to add a fixup to the _fail method so that
the retrace-log symlink is added, otherwise the task is failed and the UI
shows no 'retrace-log' link.

Finally, improve our failed task email notification to point users in the
right direction for this type of failure.


DaveWysochanskiRH:
This a small change that will fail tasks where we identify the kernel version but that crash exits with a non-zero status and the backtrace is very small, which indicates the vmcore is probably useless.  In my opinion, these tasks should be failed not declared success.  This will allow a better cleanup since these tasks will be cleaned up after 'DeleteFailedTaskAfter' which is likely to be much shorter than 'DeleteTaskAfter'.  I have been testing this on our dev server for a few days and I think there should be no unexpected side-effects.

DaveWysochanskiRH:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1232019

PR #183, opened 2018-03-28

DaveWysochanskiRH:
…files.

is hung, the timeout on the Popen.communicate will trigger an exception
and in the retrace_log (and possibly email notification) you'll see a
message like this:
 2017-10-13 18:16:49 prepare_debuginfo failed: WARNING: Possible vmcore memory corruption or damaged file - crash 'mod' command exceeded 3600 second timeout

going that route.

### DIFF
--- a/src/config/retrace-server.conf
+++ b/src/config/retrace-server.conf
@@ -152,6 +152,9 @@ EmailNotifyFrom = retrace@localhost
 # Calculate md5sum for remote resources - changeable on manager page
 CalculateMd5 = 0
 
+# Timeout (in seconds) for communication with any process
+ProcessCommunicateTimeout = 3600
+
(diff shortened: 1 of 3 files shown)


DaveWysochanskiRH:
This fixes https://bugzilla.redhat.com/show_bug.cgi?id=1208563

It is an annoying problem that keeps coming up periodically.  Someone submits a damaged vmcore and crash will spin at 100% cpu.  Right now our only recourse is to put in a ticket to have our ops (has root privs) kill the retrace-server-worker task as well as the crash task.

This patch puts a timer on crash in the first place it runs during retrace.

DaveWysochanskiRH:
I forgot to push a slightly changed commit message

PR #184, opened 2018-03-28

DaveWysochanskiRH:
…files.

is hung, the timeout on the Popen.communicate will trigger an exception.
This will fail the task and in the retrace_log (and possibly email
notification) you'll see a message like this:
 2017-10-13 18:16:49 prepare_debuginfo failed: WARNING: Possible vmcore memory corruption or damaged file - crash 'mod' command exceeded 3600 second timeout

going that route.

### DIFF
--- a/src/retrace/retrace.py
+++ b/src/retrace/retrace.py
@@ -1760,7 +1760,14 @@ class RetraceTask:
                               stdin=PIPE, stdout=PIPE, stderr=null)
         else:
             child = Popen(crash_cmd + ["-s", vmcore, vmlinux], stdin=PIPE, stdout=PIPE, stderr=STDOUT)
-        stdout = child.communicate("mod\nquit")[0]
+        try:
+            t = 3600
+            if CONFIG["ProcessCommunicateTimeout"]:
+                t = CONFIG["ProcessCommunicateTimeout"]
+            stdout = child.communicate("mod\nquit", timeout=t)[0]
+        except:
+            child.kill()
+            raise Exception, "WARNING: Possible vmcore memory corruption or damaged file - crash 'mod' command exceeded " + str(t) + " second timeout"
         if child.returncode == 1 and "el5" in kernelver.release:
             log_info("Unable to list modules but el5 detected, trying crash fixup for vmss files")
             crash_cmd.append("--machdep")
--- a/src/config/retrace-server.conf
+++ b/src/config/retrace-server.conf
@@ -152,6 +152,9 @@ EmailNotifyFrom = retrace@localhost
 # Calculate md5sum for remote resources - changeable on manager page
 CalculateMd5 = 0
 
+# Timeout (in seconds) for communication with any process
+ProcessCommunicateTimeout = 3600
+
 [archhosts]
 i386 =
 x86_64 =
--- a/src/retrace/config.py.in
+++ b/src/retrace/config.py.in
@@ -75,6 +75,7 @@ class Config(object):
             "CalculateMd5": True,
             "CaseNumberURL": "",
             "Crashi386": "",
+            "ProcessCommunicateTimeout": 3600,
         }
 
         def __getitem__(self, key):


src/retrace/retrace.py
```diff
@@ -1760,7 +1760,14 @@ def prepare_debuginfo(self, vmcore, chroot=None, kernelver=None, crash_cmd=["cra
                               stdin=PIPE, stdout=PIPE, stderr=null)
         else:
             child = Popen(crash_cmd + ["-s", vmcore, vmlinux], stdin=PIPE, stdout=PIPE, stderr=STDOUT)
-        stdout = child.communicate("mod\nquit")[0]
+        try:
+            t = 3600
+            if CONFIG["ProcessCommunicateTimeout"]:
+                t = CONFIG["ProcessCommunicateTimeout"]
+            stdout = child.communicate("mod\nquit", timeout=t)[0]
```
marusak: `TypeError: __init__() got an unexpected keyword argument 'timeout'`
`timeout` is not in Python2 subprocess.
It is, however, available in package https://pypi.python.org/pypi/subprocess32/

DaveWysochanskiRH: Weird. I had not seen this error but I test on RHEL6 mostly. Did this move in another python version?

marusak: I tried it with python2.7 on Fedora 27, RHEL7 and failed with the error I mentioned above. I also don't see it in documentation.

Have you tried it? And are you sure it does not always raise exception?  Can you check what exception it raises? I think it will be `TypeError`.

DaveWysochanskiRH: I don’t submit pull requests that are not tested. I am testing the patch (RHEL6) and I’ve never seen that exception so it must be a difference. The patch works as designed to kill a long running crash but does not affect other short running crash processes. If we need a fix up for compatibility that is fine with me.

marusak: That is very wierd. What exact version of python do you have?
Can you run this in you python interperter?
```
from subprocess import Popen, STDOUT                                                                            
child = Popen(["sleep", "10000m"], stderr=STDOUT)                               
output = child.communicate("sleep", timeout=2)  
```

DaveWysochanskiRH: $ python
Python 2.6.6 (r266:84292, Aug  9 2016, 06:11:56) 
[GCC 4.4.7 20120313 (Red Hat 4.4.7-17)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from subprocess import Popen, STDOUT
>>> child = Popen(["sleep", "1000m"], stderr=STDOUT)
>>> output = child.communicate("sleep", timeout=2)
  File "<stdin>", line 1, in <module>
  File "/usr/lib64/python2.6/subprocess.py", line 732, in communicate
    stdout, stderr = self._communicate(input, endtime)
  File "/usr/lib64/python2.6/subprocess.py", line 1352, in _communicate
    self.wait(timeout=self._remaining_time(endtime))
  File "/usr/lib64/python2.6/subprocess.py", line 1302, in wait
    raise TimeoutExpired
subprocess.TimeoutExpired

DaveWysochanskiRH:
If you are ok with the patch overall, then I think we should just fixup what is required for a proper upstream patch going forward.  I can handle fixup for RHEL6 which you said prior is not supported anymore and I hope to move off of soon anyway.  The more important thing is whether this approach is good or not and what you think of the patch.

DaveWysochanskiRH:
Maybe RHEL7 does not have this backport?

$ rpm -q python
python-2.6.6-66.el6_8.x86_64
$ rpm -q python --changelog
...

* Wed Jan 19 2011 David Malcolm <dmalcolm@redhat.com> - 2.6.6-7
- add an optional "timeout" argument to the subprocess module (patch 131)

DaveWysochanskiRH:
See https://bugzilla.redhat.com/show_bug.cgi?id=497133#c25

DaveWysochanskiRH:
It looks like we need a separate patch as you mention for upstream / rhel7 - from Nir:
"This patch is not needed since vdsm is using the python-subprocess32 package,
which is a backport of subprocess package from python 3.2 including timeouts."

DaveWysochanskiRH:
Do you want me to resubmit a rhel7 / upstream patch or are you going to fix this up?  Sorry about this I should have looked carefully at a possible API incompatibility before putting this PR in.

marusak:
We can build https://koji.fedoraproject.org/koji/packageinfo?packageID=15499 for epel7 and you fix is good to go. Otherwise It needs to implement timeout differently, e.g. https://stackoverflow.com/questions/1191374/using-module-subprocess-with-timeout 

I would start by trying to get python-subprocess32 build for epel7.

DaveWysochanskiRH:
Don't build that yet - it should not be necessary. Let me look at what vdsm does in RHEL7 per Nir note here https://bugzilla.redhat.com/show_bug.cgi?id=497133#c27

I'll check it out and resubmit an upstream patch maybe in a week or two if not sooner.

DaveWysochanskiRH:
Ok, a quick check and it looks like for el7, python-subprocess32 is only backported for RHEV channel which may mean you're right we need to build for epel7.

DaveWysochanskiRH:
I think we may be able to handle this on RHEL7 with software collections 2.x as it provides python 3.x  I am not sure if there is something analogous in Fedora we can use.  Then we can just require python 3 for retrace and use a proper timeout API.

marusak:
Migrating into python3 is a much bigger chapter than this problem. I don't think that is the solution for this issue.

trapas:

src/retrace/retrace.py
```diff
@@ -1760,7 +1760,14 @@ def prepare_debuginfo(self, vmcore, chroot=None, kernelver=None, crash_cmd=["cra
...
+        except:
+            child.kill()
+            raise Exception, "WARNING: Possible vmcore memory corruption or damaged file - crash 'mod' command exceeded " + str(t) + " second timeout"
```
mkutlak: In Python3 the syntax is:
```python3
raise Exception(...)
```

Can you also make the line fit ~120 characters max?

mkutlak:
Can you fix the file conflicts? 
retrace-server is now in Python3 so we can merge this PR.


@trapas add to whitelist

DaveWysochanskiRH:
Yes I will work on getting this fixed up and tested on our rhel8 box.  Thanks.

DaveWysochanskiRH:
Sorry I forgot to force push.  The new one is at https://github.com/abrt/retrace-server/pull/236

PR #186, opened 2018-04-16
easy requires of createrepo on RHEL

xsuchy:
There is no such createrepo >= 0.9.9-27 and therefore retrace-server is not installable on RHEL7

### DIFF
--- a/retrace-server.spec.in
+++ b/retrace-server.spec.in
@@ -24,7 +24,11 @@ Requires: p7zip
 Requires: unzip
 Requires: lzop
 Requires: elfutils
+%if 0%{?rhel} == 6 || 0%{?rhel} == 7
+Requires: createrepo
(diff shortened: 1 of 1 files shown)


marusak:
Are you sure there isn't one on RHEL7? I am pretty sure I've check it before.
Not forcing some version may easily end up that https://github.com/abrt/retrace-server/commit/20edec69f3563c5f997fe795265ad8e08f3ec2cf#diff-d74f338eb545f1389ae7ea37cfb06d39 will not work.

xsuchy:
@marusak  I hit this when I tried to upgrade retrace.fedoraproject.org. You can try it. :) If you find a suitable package somewhere, I will cancel this PR._

marusak:
Suitable package has been found, closing.

PR #187, opened 2018-04-18
For vmcores that fail crash but have a large enough kernel log, try -…

DaveWysochanskiRH:
…-minimal

A previous patch for https://bugzilla.redhat.com/show_bug.cgi?id=1232019
marked vmcores as 'failed' if the 'sys' command in crash would exit with
non-zero status, indicating this basic crash command failed, and we could
not obtain a reasonably sized (1024 bytes or above) log file.

This patch handles the case where 'sys' fails but we have a larger kernel
log file.  In that case, we at least have some place to start debug, and
so 'crash --minimal' mode seems appropriate here.

### DIFF
--- a/src/retrace/retrace_worker.py
+++ b/src/retrace/retrace_worker.py
@@ -867,9 +867,15 @@ class RetraceWorker(object):
                 crash_foreach_bt = None
 
         task.set_backtrace(kernellog)
-        # If crash sys command exited with non-zero status and log is less than 1024 bytes, probably it is not useful
-        if not crash_sys_c and len(kernellog) < 1024:
-            raise Exception("Failing task due to crash exiting with non-zero status and small kernellog size = %d bytes" % len(kernellog))
+        # If crash sys command exited with non-zero status, we likely have a semi-useful vmcore
+        if not crash_sys_c:
+            # FIXME: Probably a better hueristic can be done here
+            if len(kernellog) < 1024:
+                # If log is less than 1024 bytes, probably it is not useful at all so fail it
+               raise Exception("Failing task due to crash exiting with non-zero status and small kernellog size = %d bytes" % len(kernellog))
+            else:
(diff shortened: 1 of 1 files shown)


src/retrace/retrace_worker.py
```diff
@@ -867,9 +867,15 @@ def start_vmcore(self, custom_kernelver=None):
...
-            raise Exception("Failing task due to crash exiting with non-zero status and small kernellog size = %d bytes" % len(kernellog))
+        # If crash sys command exited with non-zero status, we likely have a semi-useful vmcore
+        if not crash_sys_c:
+            # FIXME: Probably a better hueristic can be done here
+            if len(kernellog) < 1024:
+                # If log is less than 1024 bytes, probably it is not useful at all so fail it
+               raise Exception("Failing task due to crash exiting with non-zero status and small kernellog size = %d bytes" % len(kernellog))
+            else:
+                # If log is 1024 bytes or above, try 'crash --minimal'
+                task.set_crash_cmd("crash --minimal")
```
marusak: You set crash_cmd, but is it called?

DaveWysochanskiRH: It will be anytime someone uses "retrace-server-interact <taskid> crash" for this task.
The idea is if we don't set this here someone calling that would just have crash fail to load at all.
Since 'sys' failed, we kinda know the vmcore is damaged, but we have the kernel log so it makes sense to run in minimal mode.

marusak: Oh, ok. I thought you wanted to retry with `--minimal` right away. But if the goal is to prepare it for soembody who will run `r-s-interact` then it makes sense to  me.

DaveWysochanskiRH: I should have also mentioned this is to address a backlog bug https://bugzilla.redhat.com/show_bug.cgi?id=1328227 which is not too important but I felt was worth fixing with this small patch

PR #190, opened 2018-05-11
Add bugzilla field and related tools

mkutlak:
The field is to be filled with bugzilla numbers. This should help to create
an association between task and bugzilla bugs.

The field is used by these tools:
retrace-server-bzquery:
  Query bugzilla for trigger words. For found tasks fill the bugzilla field
  with bugzilla number.

retrace-server-bugzilla-cleanup:
  Look for tasks with bugzilla field filled in and check the status
  of entered bugzilla bugs. Refresh ones with open bugzilla bugs.

### DIFF
--- a/src/manager.wsgi
+++ b/src/manager.wsgi
@@ -8,7 +8,7 @@ from retrace import *
 
 CONFIG = config.Config()
 
-MANAGER_URL_PARSER = re.compile("^(.*/manager)(/(([^/]+)(/(__custom__|start|backtrace|savenotes|caseno|notify|delete(/(sure/?)?)?|misc/([^/]+)/?)?)?)?)?$")
+MANAGER_URL_PARSER = re.compile("^(.*/manager)(/(([^/]+)(/(__custom__|start|backtrace|savenotes|caseno|bugzillano|notify|delete(/(sure/?)?)?|misc/([^/]+)/?)?)?)?)?$")
 
 LONG_TYPES = {TASK_RETRACE: "Coredump retrace",
               TASK_DEBUG: "Coredump retrace - debug",
@@ -124,6 +124,14 @@ def application(environ, start_response):
                     # caseno is invalid number - do nothing, it can be set later
                     pass
 
+            if "bugzillano" in get:
+                try:
+                    bugzillano = filter(int, set(n.strip() for n in get["bugzillano"][0].replace(";", ",").split(",")))
+                    task.set_bugzillano(bugzillano)
+                except Exception:
+                    # bugzillano is invalid number - do nothing, it can be set later
+                    pass
+
         if not task.get_managed():
             return response(start_response, "403 Forbidden", _("Task does not belong to task manager"))
 
@@ -192,6 +200,25 @@ def application(environ, start_response):
                     return response(start_response, "404 Not Found", _("Case number must be an integer; %s" % ex))
 
                 task.set_caseno(caseno)
+
+        return response(start_response, "302 Found", "", [("Location", "%s/%d" % (match.group(1), task.get_taskid()))])
+    elif match.group(6) and match.group(6) == "bugzillano":
+        POST = urlparse.parse_qs(request.body, keep_blank_values=1)
+        try:
+            task = RetraceTask(filename)
+        except Exception:
+            return response(start_response, "404 Not Found", _("There is no such task"))
+
+        if "bugzillano" in POST and len(POST["bugzillano"]) > 0:
+            if not POST["bugzillano"][0]:
+                task.delete(RetraceTask.BUGZILLANO_FILE)
+            else:
+                try:
+                    bugzillano = filter(int, set(n.strip() for n in POST["bugzillano"][0].replace(";", ",").split(",")))
+                except ValueError as ex:
+                    return response(start_response, "404 Not Found", _("Bugzilla numbers must be integers; %s" % ex))
+
+                task.set_bugzillano(bugzillano)
 
         return response(start_response, "302 Found", "", [("Location", "%s/%d" % (match.group(1), task.get_taskid()))])
     elif match.group(6) and match.group(6) == "backtrace":
@@ -307,6 +334,7 @@ def application(environ, start_response):
             startcontent = "    <form method=\"get\" action=\"%s/start\">" \
                            "      Kernel version (empty to autodetect): <input name=\"kernelver\" type=\"text\" id=\"kernelver\" /> e.g. <code>2.6.32-287.el6.x86_64</code><br />" \
                            "      Case no.: <input name=\"caseno\" type=\"text\" id=\"caseno\" /><br />" \
+                           "      Bugzilla no.: <input name=\"bugzillano\" type=\"text\" id=\"bugzillano\" /><br />" \
                            "      E-mail notification: <input name=\"notify\" type=\"text\" id=\"notify\" /><br />" \
                            "      <input type=\"checkbox\" name=\"debug\" id=\"debug\" checked=\"checked\" />Be more verbose in case of error<br />" \
                            "      <input type=\"checkbox\" name=\"md5sum\" id=\"md5sum\" %s />Calculate md5 checksum for all downloaded resources<br />" \
@@ -435,6 +463,22 @@ def application(environ, start_response):
                      "  </td>" \
                      "</tr>" % (request.path_url.rstrip("/"), currentcaseno)
 
+        bugzillano = ""
+        if not ftptask:
+            currentbugzillano = ""
+            if task.has_bugzillano():
+                currentbugzillano = "value=\"%s\"" % ", ".join(task.get_bugzillano())
+
+            bugzillano = "<tr>" \
+                     "  <th>Bugzilla no.:</th>" \
+                     "  <td>" \
+                     "    <form method=\"post\" action=\"%s/bugzillano\">" \
+                     "      <input type=\"text\" name=\"bugzillano\" %s/>" \
+                     "      <input type=\"submit\" value=\"Update bugzilla no.\" class=\"button\" />" \
+                     "    </form>" \
+                     "  </td>" \
+                     "</tr>" % (request.path_url.rstrip("/"), currentbugzillano)
+
         back = "<tr><td colspan=\"2\"><a href=\"%s\">%s</a></td></tr>" % (match.group(1), _("Back to task manager"))
 
         notes = ""
@@ -479,6 +523,7 @@ def application(environ, start_response):
         output = output.replace("{backtrace}", backtrace)
         output = output.replace("{backtracewindow}", backtracewindow)
         output = output.replace("{caseno}", caseno)
+        output = output.replace("{bugzillano}", bugzillano)
         output = output.replace("{notify}", notify)
         output = output.replace("{delete}", delete)
         output = output.replace("{delete_yesno}", delete_yesno)
@@ -550,6 +595,10 @@ def application(environ, start_response):
                         except:
                             pass
 
+                bugzillano = ""
+                if task.has_bugzillano():
+                    bugzillano = ", ".join(task.get_bugzillano())
+
                 files = ""
                 if task.has_downloaded():
                     files = task.get_downloaded()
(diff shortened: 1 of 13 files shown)


DaveWysochanskiRH:
Thank you for this patchset!  I will begin to review it either today or early next week.

src/manager.wsgi
```diff
@@ -193,6 +201,26 @@ def application(environ, start_response):
...
+            return response(start_response, "404 Not Found", _("There is no such task"))
+
+        if "bugzillano" in POST and len(POST["bugzillano"]) > 0:
+            if not POST["bugzillano"][0]:
+                task.delete(RetraceTask.BUGZILLANO_FILE)
+            else:
+                try:
+                    bugzillano = filter(int, set(n.strip() for n in POST["bugzillano"][0].replace(";", ",").split(",")))
+                    # int(POST["bugzillano"][0])
+                except Exception as ex:
```
xsuchy: again. No need to catch generic exception.

src/retrace-server-bugzilla-cleanup
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/python2
+import bugzilla
+from retrace import *
```
xsuchy: Can you enumerate what are you importing?

src/retrace-server-bugzilla-cleanup
```diff
@@ -0,0 +1,55 @@
...
+        try:
+            files = os.listdir(CONFIG["SaveDir"])
+        except OSError as ex:
+            files = []
+            log.write("Error listing task directory: %s\n" % ex)
+
+        for filename in files:
+            try:
+                task = RetraceTask(filename)
+            except:
```
xsuchy: too general exception

src/retrace-server-bzquery
```diff
@@ -0,0 +1,109 @@
+#!/usr/bin/python2
+import bugzilla
+from retrace import *
```
xsuchy: can you enumerate what are you importing?

DaveWysochanskiRH:
bzquery-tool

While testing, I found this tool cannot be run as a normal user (which is good) because you get this
$ retrace-server-bzquery 
  File "/usr/bin/retrace-server-bzquery", line 11, in <module>
    with open(logfile, "a") as log:
IOError: [Errno 13] Permission denied: '/var/log/retrace-server/bzquery.log'

But unfortunately if the tool is run as root (I think the intent is run from a cronjob yes?), then the 'bugzillano' files will be owned by root which looks incorrect (should be owned by 'retrace' user:
 ls -lh /retrace/tasks/748692525/
total 1.1M
-rw-r--r--. 1 root    retrace     8 May 11 11:38 bugzillano
-rw-r--r--. 1 retrace retrace     7 Oct 20  2017 caseno
drwxrwx---. 2 retrace retrace  4.0K Oct 20  2017 crash
...

I don't think it needs root perms since it's not removing anything so should the tool run as retrace user?  What is the intent?

DaveWysochanskiRH:
naming inconsistency with tools: retrace-server-bzquery vs retrace-server-bugzilla-cleanup
probably just pick one - personally the longer one is clearer to me so suggest renaming 'retrace-server-bzquery' to 'retrace-server-bugzilla-query'

DaveWysochanskiRH:
Testing - a bit difficult to test outside of production machine, but here's what I did.  From what I can tell this works as designed so well done!  So far I only found minor bugs (in other feedback comments).

1. Copied some task directories from production machine to development / test machine and did the rest of the steps from this machine

2. ran retrace-server-bzquery as root
  - verified the tasks had the bugzillano field filled in properly
  - verified no other tasks had bugzillano field
  - verified the bugzilla number inside bugzillano was the smallest numbered bz for a task in multiple bzs

3. ran retrace-server-bugzilla-cleanup as root
  - verified this tool updated the mtime/ctime of the tasks with a bugzillano field filled in and the bugzilla is in 'ASSIGNED' state
  - verified this tool did not update the mtime/ctime of a task with a bugzillano field filled in a a bugzilla in 'CLOSED' state
  - verified this tool did not update the mtime/ctime of tasks without a bugzillano field

4. verified the 'manager' web UI works as expected
  - some tasks have 'bugzillano' filled in, others do not
  - able to manually set the 'bugzillano' field when navigating to a specific task and clicking 'update'

5. check both bugzilla-cleanup.log and bzquery.log
  - both logs reflected a record of the tool execution

mkutlak:
Changes I've made:
- Changed the default bugzilla status config to suggested one: `VERIFIED, RELEASE_PENDING, CLOSED`
- Renamed `r-s-bzquery` to `r-s-bugzilla-query`
- Query remaining bugzilla statuses
- Added reset age method
- Few bugs/errors in code I found while I was making these changes

DaveWysochanskiRH:
Thanks for the updates!  I will take a look and update my test machine.

DaveWysochanskiRH:
Is there any easy way to store these in a non-world readable file?
+BugzillaUsername =
+# Password to log in to Bugzilla (if needed)
+BugzillaPassword =

Today unfortunately we store all credentials (for example ftp) in /etc/retrace-server.conf which is world-readable.  Now we are adding a second credential and this is potentially much worse due to what it is accessing.  I am concerned about adding this due to the bugzilla capabilities this account will need to do the job well.

mkutlak:
> Is there any easy way to store these in a non-world readable file?
> +BugzillaUsername =
> +# Password to log in to Bugzilla (if needed)
> +BugzillaPassword =

I thought about adding an option `--credentials` that would load file with stored credentials. Does that sound good to you?

BTW, you can use `RETRACE_SERVER_CONFIG_PATH` to set a different path to the configuration file.

DaveWysochanskiRH:
Yes the —credentials option to the command should work ok. It should be run as root I think so we just need to figure out the right place for that file. The problem with the configure file is it has to be world readable for retrace to work. It’s possible there’s a small bug there and if that was fixed then we could just restrict permissions. However I haven’t looked into that so I’m not sure the extent of the problem.

DaveWysochanskiRH:
I think probably the conf file has to be world readable because of the retrace-server-* commands read various CONFIG variables and also call into retrace and retrace_worker classes which also use more variables. So at least at first glance the world readable nature of Configure file seems designed in and probably is not easy to fix. So I think the best option at least for short term may be some —credentials option on the command. Note that if you use the “bugzilla” command line from python-bugzilla there is a “login” command and this uses some other credential file once logged in (user doesn’t not have to store password in file system). That may be another option.

DaveWysochanskiRH:
Thanks for the quick turnaround!  I will get this into a build and start testing, hopefully today.

DaveWysochanskiRH:
manager: Make the bugzillano a clickable link

There's something wrong with this patch.  When I added this in it is not displaying anything in the bugzilla 'manager' column anymore.

DaveWysochanskiRH:
Yeah for some reason the min(task.get_bugzillano()) is returning no value or at least the output is showing no bug number.  Here is what my browser is printing as far as html

<td><a href="https://bugzilla.redhat.com/"></a></td>

This is for a row where the bugzillano file contains a single bug:
 cat /retrace/tasks/748692525/bugzillano 
1448750

DaveWysochanskiRH:
Ok if I change this to just 
~~~
bugzillano = task.get_bugzillano()
~~~

I get this in the browser html
~~~
<td><a href="https://bugzilla.redhat.com/set%28[%27%27," '1448750'])="">set(['', '1448750'])</a></td>
~~~

DaveWysochanskiRH:
Ok I figured it out.  The difference between 'caseno' and 'bugzillano' is that 'bugzillano' there is a newline at the end.  So I think there may be a problem with set_bugzillano - we need to not write a newline or handle the newline so it does not write in a '' entry into the set.  This fixes the problem for example:
~~~
echo -n "1448750" > /retrace/tasks/748692525/bugzillano
~~~

mkutlak:
> Ok I figured it out. ...

Nice, I will fix this.

mkutlak:
It should be fixed now.

DaveWysochanskiRH:
Actually I think I do need a fixup to handle the various RHEL releases at least since there's still 20 tasks in production with open RHEL6 bugs.  So we need to cover that somehow - either with a parameter to the query script or a CONFIG variable (preferable).  I ran this against production dataset (by copying the task directories to another test retrace-server instance) and it ran fine.  I think the patchset will be great with this one last fixup.

DaveWysochanskiRH:
Thanks - I will rebuild and test this!

DaveWysochanskiRH:
I think something is broken in retrace-server-bugzilla-query as I'm getting no tasks with bugzillas associated.  Will work on debugging.

DaveWysochanskiRH:
I would suggest calling these "BugzillaTriggerWords" and "BugzillaRegExes" respectively to be consistent with the other variables.

Also you may want to consider another name for 'retrace-server-bugzilla-cleanup' as it is not really doing any cleanup but refreshing tasks that are associated with bugzillas.  Possibly 'retrace-server-bugzilla-refresh' tells a little more closely what the command does, or even 'retrace-server-bugzilla-task-refresh' but that is a little long.  Maybe you can think of something else?

FWIW, originally I had a problem due to the move to the config file for the TriggerWords and RegExes.  I had neglected to change the root directory for the tasks as well as the '\\n' was causing a problem in the regex.  I am using the following which seems to work fine:
~~~
BugzillaProduct = Red Hat Enterprise Linux 6, Red Hat Enterprise Linux 7, Red Hat Enterprise Linux 8
BugzillaComponent = kernel
TriggerWords = retrace-server-interact, retrace/tasks
RegExes = retrace-server-interact\s+([0-9]{9}), /cores/retrace/tasks/([0-9]{9})/crash/vmcore
~~~

I am going to do a bit more testing with retrace-server-task command and will give final feedback.

DaveWysochanskiRH:
I think the patchset is good to merge (though consider the renames in previous comment).  Some final tests:

retrace-server-bugzilla-query
* ensure tasks with associated open bugs on RHEL6 and RHEL7 have a bugzillano filled in: PASS
* ensure task with 'bugzillano' filled in but in CLOSED state is not seen in log

retrace-server-bugzilla-cleanup
* bugzillano filled in, but invalid #: PASS (time of task is not updated)
* bugzillano filled in, CLOSED: PASS (time of task not updated)

retrace-server-task:
* 'get' --bugzillano (with 1 or two numeric values set): PASS
* 'get' --bugzillano (with NULL value set): PASS (printed 'Bugzilla number is not set.')
- 'set' number: PASS
- 'set' number1, number2: PASS

'manager' page:
- 'bugzillano' delete field: PASS 
- 'bugzillano' set field currently empty: PASS
- bugzillano' set field previously not-empty: PASS

NOTE: I've got the following root crontab - probably the bugzilla* commands only need to be run once per week so in production I'll change those.
~~~
0 * * * * /usr/bin/retrace-server-cleanup >> /var/log/retrace-server/cleanup_error.log 2>&1
20 * * * * /usr/bin/retrace-server-bugzilla-query >> /var/log/retrace-server/bugzilla_query_error.log 2>&1
40 * * * * /usr/bin/retrace-server-bugzilla-cleanup >> /var/log/retrace-server/bugzilla_cleanup_error.log 2>&1
~~~

DaveWysochanskiRH:
Thanks!  Will take a look.

DaveWysochanskiRH:
A couple small nits.

1. In "Implement bugzilla field", patch header still refers to the 'retrace-server-bzquery' and 'retrace-server-bugzilla-cleanup' - "git rebase -i" is your friend. ;-)

2. In "Implement retrace-server-bugzilla-refresh tool", rename the log file:
+    logfile = os.path.join(CONFIG["LogDir"], "bugzilla-cleanup.log")

I have the latest code running in a RHEL6 and RHEL7 build and so far looks good.

mkutlak:
> A couple small nits.

Fixed. Thank you. :)

DaveWysochanskiRH:
I am running the latest code and it looks good - clear to merge as far as I'm concerned.

src/manager.wsgi
```diff
@@ -124,6 +124,14 @@ def application(environ, start_response):
                     # caseno is invalid number - do nothing, it can be set later
                     pass
 
+            if "bugzillano" in get:
+                try:
+                    bugzillano = filter(int, set(n.strip() for n in get["bugzillano"][0].replace(";", ",").split(",")))
+                    task.set_bugzillano(bugzillano)
+                except Exception:
```
xsuchy: Can this be a specific exception?

DaveWysochanskiRH: If you have ideas of how to trip exceptions like this let me know and I can try to test them. I'm not sure exactly what this will catch as the input methods seem protected.  For example, I tried to put a non-number in via the 'manager' web UI and it would not take it.  I also cannot trigger this via retrace-server-task:

$ retrace-server-task set --no-verify --bugzillano 11542602,fred 148364915
Status code 404 was returned with message: 'Bugzilla numbers must be integers; invalid literal for int() with base 10: 'fred''

PR #205, opened 2018-06-05
py3 compatibility: Adjust raise statement syntax

ari3s:
The syntax for raising exceptions changed between Python 2 and Python 3.


mkutlak:
LGTM.

PR #206, opened 2018-06-05
py3 compatibility: Adjust imports

ari3s:
In Python 3, imports can be either absolute or explicitly relative; implicit relative imports are forbidden.

### DIFF
--- a/src/retrace/retrace.py
+++ b/src/retrace/retrace.py
@@ -19,11 +19,11 @@ import hashlib
 from signal import *
 from subprocess import *
 import magic
-from argparser import *
+from .argparser import *
 from webob import Request
 from yum import YumBase
-from config import *
-from plugins import *
+from .config import *
+from .plugins import *
 from rpmUtils.miscutils import splitFilename
 import six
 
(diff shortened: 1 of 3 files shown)


src/retrace/retrace.py
```diff
@@ -19,11 +19,11 @@
 from signal import *
 from subprocess import *
 import magic
-from argparser import *
+from .argparser import *
 from webob import Request
 from yum import YumBase
 from config import *
```
mkutlak: Missed this one.

ari3s: @mkutlak , thank you for the review. However, I do not understand your comment. May I ask you clarify more?

mkutlak: `from config import *` imports config options from [config.py.in](https://github.com/abrt/retrace-server/blob/master/src/retrace/config.py.in), therefore it should be 
`from .config import *`, or am I wrong?

ari3s: @mkutlak I see now, you are right, thanks. I update the patch.

PR #207, opened 2018-06-05
py3 compatibility: Resolution of range and xrange

ari3s:
In Python 2, `range()` returns a list, and `xrange()` returns an iterator. In Python 3, `range()` returns an iterator, and `xrange()` doesn’t exist. The compatibility is solved by `six` library.

### DIFF
--- a/src/retrace/retrace.py
+++ b/src/retrace/retrace.py
@@ -26,6 +26,7 @@ from config import *
 from plugins import *
 from rpmUtils.miscutils import splitFilename
 import six
+from six.moves import range
 
 GETTEXT_DOMAIN = "retrace-server"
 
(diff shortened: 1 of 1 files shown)


src/retrace-server-task
```diff
@@ -18,7 +18,7 @@ from requests_kerberos import HTTPKerberosAuth, OPTIONAL
 
 
 TASK_RETRACE, TASK_DEBUG, TASK_VMCORE, TASK_RETRACE_INTERACTIVE, \
-        TASK_VMCORE_INTERACTIVE = range(5)
+        TASK_VMCORE_INTERACTIVE = list(range(5))
```
mkutlak: The `list()` is not needed here.

PR #209, opened 2018-06-06
py3 compatibility: Replace filter function with a list equivalent

ari3s:
In Python 2, `filter(function, iterable)` returns a list and is equivalent
to `[item for item in iterable if function(item)]` if function is not `None`
and `[item for item in iterable if item]` if function is `None`. In Python 3,
the function returns an iterator. The replacement of Python 2 `filter()`
function with a proper equivalent or wrapping a call to `filter()`
with a call to `list()` solves the Python 3 compatibility.

### DIFF
--- a/src/retrace/retrace.py
+++ b/src/retrace/retrace.py
@@ -1653,7 +1653,7 @@ class RetraceTask:
 
     def get_notify(self):
         result = self.get(RetraceTask.NOTIFY_FILE, maxlen=1 << 16)
-        return filter(None, set(n.strip() for n in result.split("\n")))
+        return [email for email in set(n.strip() for n in result.split("\n")) if email]
 
     def set_notify(self, values):
         if not isinstance(values, list) or not all([isinstance(v, six.string_types) for v in values]):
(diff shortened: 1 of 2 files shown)


src/retrace/retrace.py
```diff
@@ -1653,7 +1653,7 @@ def has_notify(self):
 
     def get_notify(self):
         result = self.get(RetraceTask.NOTIFY_FILE, maxlen=1 << 16)
-        return filter(None, set(n.strip() for n in result.split("\n")))
+        return [_f for _f in set(n.strip() for n in result.split("\n")) if _f]
```
mkutlak: The `get_notify` loads e-mails (for notifications) from file.

Can you change the `_f` to `mail` or `email`?

src/retrace/retrace.py
```diff
@@ -2300,7 +2300,7 @@ def get_bugzillano(self):
         if result is None:
             return None
 
-        return filter(None, set(n.strip() for n in result.split("\n")))
+        return [_f for _f in set(n.strip() for n in result.split("\n")) if _f]
```
mkutlak: The `get_bugzillano` loads bugzilla numbers from file.

Can you change the `_f` to `bznum` or something similar?

ari3s: @mkutlak Thanks for the review. I update the patch.

PR #213, opened 2018-06-18
do not require install-info on F28+

xsuchy:
See
https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/MP2QVJZBOJZEOQO2G7UB2HLXKXYPF2G5/


xsuchy:
retest this please

PR #216, opened 2018-06-21
Skip tasks with no vmcore or invalid md5sum

DaveWysochanskiRH:
I did not file any bug for these two changes but they are minor fixups related to the md5sum patchset to cover a few corner cases.  This was tested in our development / test retrace-server instance and avoids the problems described in the patch header and not causing any valid dedup to be skipped.

### DIFF
--- a/src/retrace/retrace.py
+++ b/src/retrace/retrace.py
@@ -934,7 +934,7 @@ def get_md5_tasks():
         if not task.has_finished_time():
             continue
 
-        if not task.has_vmcore():
+        if not task.has_vmcore() and not task.has_coredump():
             continue
 
         if not task.has_md5sum():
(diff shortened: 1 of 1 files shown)


trapas:

marusak:
test this please

marusak:
add to whitelist

DaveWysochanskiRH:
I'm not sure I understand these comments - were these directed at me or someone else?

marusak:
@DaveWysochanskiRH those were for our bot :) sorry for the confusion, I remember when someone commented for the first time `test this please` on my PR I got confused, as I tested it and did not understand why he wants me to test it again.

src/retrace/retrace.py
```diff
@@ -1678,6 +1690,10 @@ def get_vmlinux(self):
     def set_vmlinux(self, value):
         self.set(RetraceTask.VMLINUX_FILE, value)
 
+    def has_vmcore(self):
```
marusak: Could you make this work with coredump as well?
It is the same as `vmcore`, but is called `coredump`.  Maybe rename function to `has_core` and check both, or when you check that if `has_vmcore` check if `has_vmcore or has_coredump`.

DaveWysochanskiRH: I think a 'has_coredump' is better.  Unfortunately I do not have the ability to test userspace cores because that has never worked correctly on our instances but I can put another patch on this series.

PR #224, opened 2019-02-10
cleanup: Add check for opened crash files

mkutlak:
Adds a new check to Delete(Failed)TaskAfter procedure to prevent tasks with opened/used vmcore/coredump files from being deleted.

Related: rhbz#1253908

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -101,6 +101,15 @@ AC_PATH_PROG([XMLTO], [xmlto], [no])
     [exit 1]
 [fi]
 
+AC_PATH_PROG([LSOF], [lsof], [no])
+[if test "$LSOF" = "no"]
+[then]
+    [echo "The lsof program was not found in the search path. Please ensure"]
+    [echo "that it is installed and its directory is included in the search path."]
+    [echo "Then run configure again before attempting to build Retrace server."]
+    [exit 1]
+[fi]
+
 PKG_PROG_PKG_CONFIG
 AC_ARG_WITH([systemdsystemunitdir],
         AS_HELP_STRING([--with-systemdsystemunitdir=DIR], [Directory for systemd service files]),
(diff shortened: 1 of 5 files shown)


DaveWysochanskiRH:
It looks pretty good - I'm going to build and do a few tests.  Thanks for the quick turnaround!

DaveWysochanskiRH:
This works but one possible fixup / problem to consider.  What if 'lsof' is not installed?

Other commands such as "df" are handled like this:
src/retrace/config.py.in:DF_BIN = "@DF_BIN@"
src/retrace/Makefile.am:	sed -e "s|@DF_BIN@|$(DF)|g" \
src/retrace/retrace.py:    child = Popen([DF_BIN, "-B", "1", path], stdout=PIPE, encoding='utf-8')

If we just avoid checking in any way, we get an error in cleanup log like this (NOTE: I ran this on rhel6 so I removed the 'encoding' argument from the Popen, but that is all I changed):

==> /var/log/retrace-server/cleanup_error.log <==
  File "/usr/bin/retrace-server-cleanup", line 189, in <module>
    if check_open_crash_file(task):
  File "/usr/bin/retrace-server-cleanup", line 46, in check_open_crash_file
    lsof = Popen(["lsof", "+wt", crash_path], stdout=PIPE)
  File "/usr/lib64/python2.6/subprocess.py", line 642, in __init__
    errread, errwrite)
  File "/usr/lib64/python2.6/subprocess.py", line 1238, in _execute_child
    raise child_exception
OSError: [Errno 2] No such file or directory

mkutlak:
Thank for pointing that out. 

I've added `lsof` to Requires in spec file and changed the handling to be consistent with other commands (`df`, `du`, ...).

DaveWysochanskiRH:
Commit 7c3499c looks good to me on rhel6 and rhel7.  Thanks!

PR #227, opened 2019-03-15
Fix get_kernel_release() on Python3

DaveWysochanskiRH:
The get_kernel_release() function reads binary data and tries to match
regexp strings to find kernel versions.  For Python3 we need the regexps
to thus be byte strings.  Once we've found some match, we're down to
a small character set and we can convert that to a string and strip
any spaces, newline or return characters before returning.

This fixes retrace-server task failures with log entries that look like
this:
2019-03-15 14:09:37 prepare_debuginfo failed: cannot use a string pattern on a bytes-like object
    2019-03-15 14:09:37 Stripped size: 32.01 GB
    2019-03-15 14:09:37 Makedumpfile took 0 seconds and saved 0.00 B
    2019-03-15 14:09:37 Analyzing crash data
    2019-03-15 14:09:37 cannot use a string pattern on a bytes-like object

### DIFF
--- a/src/retrace/retrace.py
+++ b/src/retrace/retrace.py
@@ -521,7 +521,7 @@ def is_package_known(package_nvr, arch, releaseid=None):
 # 1. Look for 'OSRELEASE='.  For example:
 # OSRELEASE=2.6.18-406.el5
 # NOTE: We can get "OSRELEASE=%" so we disallow the '%' character after the '='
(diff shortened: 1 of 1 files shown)


DaveWysochanskiRH:
This has got another problem so don't merge yet.

DaveWysochanskiRH:
This should be ready to merge

mkutlak:
:+1:

Thank you

PR #228, opened 2019-03-20
Fix ftp.wsgi for Python3 due to sorted() method api change, speed man…

DaveWysochanskiRH:
…ager load

In Python3 the sorted() method changed to no longer have a 'cmp' method.
Further, the 'cmp' method is stated to be slow, and we've had issues
where the 'manager' page can be slow to load.  Per Python2 reference:
https://docs.python.org/2/library/functions.html#sorted
"In general, the key and reverse conversion processes are much
faster than specifying an equivalent cmp function. This is because
cmp is called multiple times for each list element while key and
reverse touch each element only once. Use functools.cmp_to_key()
to convert an old-style cmp function to a key function."

Change this to a 'key' method with equivalent functionality.  The only purpose
on the 'manager' page appears to list files with 'vmcore' in them first.

### DIFF
--- a/src/ftp.wsgi
+++ b/src/ftp.wsgi
@@ -26,7 +26,7 @@ def async_ftp_list_dir(filterexp):
     if filterexp:
         tasklist = sorted(fnmatch.filter(rawtasklist, filterexp))
     else:
-        tasklist = sorted(rawtasklist, cmp=cmp_vmcores_first)
+        tasklist = sorted(rawtasklist)
 
     for fname in tasklist:
         available.append("<tr><td><a href=\"manager/%s\">%s</a></td></tr>" \
--- a/src/retrace/retrace.py
+++ b/src/retrace/retrace.py
@@ -1189,17 +1189,6 @@ def ftp_list_dir(ftpdir="/", ftp=None):
 
     return result
 
-def cmp_vmcores_first(str1, str2):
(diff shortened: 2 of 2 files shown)


DaveWysochanskiRH:
FWIW, I tested this on RHEL8 and by backporting to RHEL6 so it works on python3 and python2.

mkutlak:
Thank you!

This should also sort the files by their name (`vmcore` are listed first). 
``` python3
def key_contains_vmcore(string):
    if "vmcore" in string.lower():
        return string.upper()
    else:
        return string.lower()
```

DaveWysochanskiRH:
The patch was just trying to fix the bug not change the sorting method.  However, since you mentioned it...

We could sort by name...
Or I considered removing the sorting altogether since:
a) it's just overhead (causes manager page to load slower, especially if there's a lot of files on ftp server)
b) most people will probably just hit CTRL-F and search for their filename anyway

What do you think Martin?  Sort by name or just remove the sort altogether?

mkutlak:
I am ok with removing of the sorting altogether.

DaveWysochanskiRH:
Tested the new commit c478fb2 on both rhel8 (python3) and rhel6 (python2) and works fine.

mkutlak:
Thank you @DaveWysochanskiRH ! :+1:

PR #239, opened 2019-05-06

DaveWysochanskiRH:
…files.

is hung, the timeout on the Popen.communicate will trigger an exception.
This will fail the task and in the retrace_log (and possibly email
notification) you'll see a message like this:
 prepare_debuginfo failed: WARNING: crash 'mod' command exceeded 3600 second timeout - damaged vmcore?

going that route.

### DIFF
--- a/src/retrace/retrace.py
+++ b/src/retrace/retrace.py
@@ -1839,7 +1839,14 @@ class RetraceTask:
         else:
             child = Popen(crash_cmd + ["-s", vmcore, vmlinux], stdin=PIPE, stdout=PIPE, stderr=STDOUT,
                           encoding='utf-8')
-        stdout = child.communicate("mod\nquit")[0]
+        try:
+            t = 3600
+            if CONFIG["ProcessCommunicateTimeout"]:
+                t = CONFIG["ProcessCommunicateTimeout"]
+            stdout = child.communicate("mod\nquit", timeout=t)[0]
+        except TimeoutExpired:
+            child.kill()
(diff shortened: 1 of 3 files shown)


DaveWysochanskiRH:
The patch is identical to what was tested in https://github.com/abrt/retrace-server/pull/236 but I fixed the commit message to show the proper log message you'll get.  This will fit inside 120 characters as requested on original PR https://github.com/abrt/retrace-server/pull/184

I apologize for the multiple PRs.  Unfortunately I don't see a way to delete the duplicates.

src/retrace/retrace.py
```diff
@@ -1839,7 +1839,14 @@ def prepare_debuginfo(self, vmcore, chroot=None, kernelver=None, crash_cmd=["cra
         else:
             child = Popen(crash_cmd + ["-s", vmcore, vmlinux], stdin=PIPE, stdout=PIPE, stderr=STDOUT,
                           encoding='utf-8')
-        stdout = child.communicate("mod\nquit")[0]
+        try:
+            t = 3600
+            if CONFIG["ProcessCommunicateTimeout"]:
+                t = CONFIG["ProcessCommunicateTimeout"]
+            stdout = child.communicate("mod\nquit", timeout=t)[0]
+        except:
```
mkutlak: Can you use a specific exception (`TimeoutExpired`) here?

DaveWysochanskiRH: Sure can!

mkutlak:
LGTM :+1: 

Thank you @DaveWysochanskiRH

PR #241, opened 2019-06-10
Use ISO-8859-1 rather than utf-8 for crash commands to avoid prematur…

DaveWysochanskiRH:
…e task failure

Some vmcores are partially damaged but still very usable.  In one case
the 'foreach bt' command returned some non-ascii characters for one
task with a corrupted stack.  The vmcore was quite usable otherwise
but retrace_log failed with a cryptic error:
2019-06-10 10:43:42 'utf-8' codec can't decode byte 0xac in position 10472: invalid start byte

The above failure did not happen previously with Python2 but now happens
on Python3.  To return to the previous behavior, just use ISO-8859-1
character set that covers these non-ascii characters when issuing any
crash command.

### DIFF
--- a/src/retrace/retrace.py
+++ b/src/retrace/retrace.py
@@ -537,7 +537,7 @@ def get_kernel_release(vmcore, crash_cmd=["crash"]):
     # set SIGPIPE to default handler for bz 1540253
     save = getsignal(SIGPIPE)
     signal(SIGPIPE, SIG_DFL)
-    child = Popen(crash_cmd + ["--osrelease", vmcore], stdout=PIPE, stderr=STDOUT, encoding='utf-8')
+    child = Popen(crash_cmd + ["--osrelease", vmcore], stdout=PIPE, stderr=STDOUT, encoding='ISO-8859-1')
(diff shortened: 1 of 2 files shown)


DaveWysochanskiRH:
FWIW, I thought of a more elaborate approach to this as it would be useful to know if some crash command output produced some non-ascii characters.  I suppose we could wrap any Popen / child.communicate and then try .. except with 'utf-8'.  If that fails print and then re-try with ISO-8859-1.

Also at various points I've thought about factoring out all the more extensive crash commands that run once we have the proper debuginfo.  These would then go into something like a default post retrace hook, with maybe a status of each of these.

Either of the above would be more extensive than the above simple patch, and I'm inclined to avoid tackling at least the full refactoring due to this problem.  However if it is desired myself or someone else could at least try the wrap of Popen which may not be too bad.

DaveWysochanskiRH:
Hold this - I'm going to try a refactoring since it may be a lot cleaner.

DaveWysochanskiRH:
I'm closing this in favor of the approach in PR #242

PR #242, opened 2019-06-11
Refactor crash commands run after prepare_debuginfo into run_crash_cm…

DaveWysochanskiRH:
…dline helper

When processing a vmcore and after prepare_debuginfo() runs successfully,
we run a series of crash commands.  These commands are the same whether
mock is used or not.  Create a helper function, run_crash_cmdline which
removes a lot of redundancy and improve maintainability.  In addition
to refactoring and maintenance improvements, there are a couple slight
functional changes with this patch.

First, the patch changes what happens with crash commands when the
'if child.wait()' condition is true.  Prior to this patch, we would call
log_warn() then set the output to 'None', with the only exception being
the 'log' command which we would not set output to 'None'.  With this
patch, we never set the output to 'None' for any command.  This should
not be an issue for most commands, and might even be viewed as a feature
since it will create output files in 'misc' for these other commands,
possibly with just error output in them.  However there is one possible
issue with creating error output in these files and involves 'foreach_bt'.
In the case of error output in 'foreach_bt', probably we should skip
calling bt_filter.  To avoid calling bt_filter, add a len() check similar
to what we do with the kernellog - if it is less than 1024 bytes we
assume it is not valid and skip calling bt_filter.

Second, the new function should handle non-printable characters in the
output of any crash command.  Prior to this patch, a partially damaged
vmcore with one or more corrupted task stacks could exit with a cryptic
error and task failure:
2019-06-10 10:43:42 'utf-8' codec can't decode byte 0xac in position 10472: invalid start byte
This only started happening on Python3 (previous python2 code handles
the same vmcore fine and does not fail the task).  This patch handles
this case by wrapping the child.communicate() in a try...except and
printing the error in that case:
 2019-06-11 17:57:51 crash command: 'set hex; foreach bt; quit; ' triggered UnicodeDecodeError
 2019-06-11 17:57:51   'utf-8' codec can't decode byte 0xac in position 10522: invalid start byte
If we hit the above error, we do not obtain any output in the command that
failed, so it does slightly change the behavior from Python2.  However,
at least now the specific command is now flagged as having an error,
the task is not failed, and the user may decide what to do with the error.

### DIFF
--- a/src/retrace/retrace_worker.py
+++ b/src/retrace/retrace_worker.py
@@ -641,6 +641,32 @@ class RetraceWorker(object):
 
         return s1.st_size
 
+    def run_crash_cmdline(self, crash_start, crash_cmdline):
+        cmd_output = None
+        returncode = 0
+        try:
+            child = Popen(crash_start, stdin=PIPE, stdout=PIPE, stderr=STDOUT, encoding='utf-8')
+            cmd_output = child.communicate(crash_cmdline)[0]
+        except OSError as err:
+            log_warn("crash command: '%s' triggered OSError " %
+                     crash_cmdline.replace('\r', '; ').replace('\n', '; '))
+            log_warn("  %s" % err)
+        except UnicodeDecodeError as err:
+            log_warn("crash command: '%s' triggered UnicodeDecodeError " %
+                     crash_cmdline.replace('\r', '; ').replace('\n', '; '))
+            log_warn("  %s" % err)
+        except:
+            log_warn("crash command: '%s' triggered Unknown exception %s" %
+                     crash_cmdline.replace('\r', '; ').replace('\n', '; '))
+            log_warn("  %s" % sys.exc_info()[0])
+
+        if child.wait():
+            log_warn("crash '%s' exitted with %d" % (crash_cmdline.replace('\r', '; ').replace('\n', '; '),
+                     child.returncode))
+            returncode = child.returncode
+
+        return cmd_output, returncode
+
     def start_vmcore(self, custom_kernelver=None):
         self.hook_start()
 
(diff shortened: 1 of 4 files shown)


DaveWysochanskiRH:
While more invasive, I like this approach better since it consolidates a lot of code.
I've tested both the mock and non-mock case, and some failure cases.
I like the fact this flags commands that fail to decode but does not fail the task, though we could just simplify run_crash_cmd() to use 'ISO-8859-1' encoding and avoid the error like previous retrace-server did (python2).

src/retrace/retrace_worker.py
```diff
@@ -641,6 +641,32 @@ def dedup_vmcore(self, md5_task):
 
         return s1.st_size
 
+    def run_crash_cmdline(self, crash_start, crash_cmdline):
+        cmd_output = None
+        returncode = 0
+        try:
+            child = Popen(crash_start, stdin=PIPE, stdout=PIPE, stderr=STDOUT, encoding='utf-8')
```
mkutlak: Could't the `encoding` be removed altogether in **this** case? It would get rid of of the decoding error.

_`add_misc()` and `set_atomic()` functions would need to be modified to work with bytes._

DaveWysochanskiRH: Personally I don't think that's a good idea since that would make this inconsistent with other calls to Popen in the code.  The way retrace-server is written today, we'll run into this problem on other binaries and commands too if called via Popen with the 'utf-8' encoding and we get some unexpected characters.

Maybe you want to re-think the approach to the original commit that added the 'utf-8' encoding, or add some exception handling?


commit e80d469087ed1118a629bbcaa3d32207b58c8646
Author: Martin Kutlak <mkutlak@redhat.com>
Date:   Fri Jul 27 14:23:47 2018 +0200

    py3-comp: Specify encoding for Popen
    
    If encoding or errors are specified, or text is true, the file objects stdin, stdout and stderr
    are opened in text mode with the specified encoding and errors. The universal_newlines argument
    is equivalent to text and is provided for backwards compatibility. [1]
    
    By default, file objects are opened in binary mode.
    
    encoding argument was added in version 3.6.
    
    [1] https://docs.python.org/3/library/subprocess.html#popen-constructor
    
mkutlak: > Personally I don't think that's a good idea since that would make this inconsistent with other calls to Popen in the code.

I think consistency isn't important with `Popen()` calls.

> Maybe you want to re-think the approach to the original commit that added the 'utf-8' encoding, or add some exception handling?

Yes, it would be good idea to go through the `Popen()` calls and modify cases where the `encoding` isn't needed. 

___
I still think we can remove the `encoding` here. 

As I see it, we run the `crash` command and save its output to a file. There is no string processing involved (no `strip()` or `split()`), so we do not need to decode it to `utf-8`.

DaveWysochanskiRH: Ok let me work on that and I'll do some testing and "force push" a new commit.

DaveWysochanskiRH: Ok so if you want me to do this then I need to change 'add_misc" as well.  Which means I also need to modify run_gdb() because it too uses 'utf-8' in Popen.  I cannot test this unfortunately can you test it?  If so I can put a more extensive patch series together of multiple patches.

mkutlak: >  I cannot test this unfortunately can you test it?

Yes, I can test it.

mkutlak: > Ok so if you want me to do this then I need to change 'add_misc" as well.

You also need to change `set_atomic` and `set_backtrace` functions (because of kernellog).

DaveWysochanskiRH: Ok I will add a few patches on top of this one because I think that makes the most sense due to all the refactoring.

DaveWysochanskiRH: When I remove 'encoding' from Popen I get a TypeError thrown inside 'run_crash_cmdline'
 2019-06-26 13:37:59 crash command: 'log; quit; ' triggered TypeError 
 2019-06-26 13:37:59   memoryview: a bytes-like object is required, not 'str'

This is the code I"m testing:
    def run_crash_cmdline(self, crash_start, crash_cmdline):
        cmd_output = None
        returncode = 0
        try:
            child = Popen(crash_start, stdin=PIPE, stdout=PIPE, stderr=STDOUT)
            cmd_output = child.communicate(crash_cmdline)[0]
        except OSError as err:
            log_warn("crash command: '%s' triggered OSError " %
                     crash_cmdline.replace('\r', '; ').replace('\n', '; '))
            log_warn("  %s" % err)
        except TypeError as err:
            log_warn("crash command: '%s' triggered TypeError " %
                     crash_cmdline.replace('\r', '; ').replace('\n', '; '))
            log_warn("  %s" % err)
        except:
            log_warn("crash command: '%s' triggered Unknown exception %s" %
                     (crash_cmdline.replace('\r', '; ').replace('\n', '; '),
                      sys.exc_info()[0]))

I'm not sure changing to binary opening of stdin/stdout/stderr is the correct approach which is what happens when we remove 'encoding' from Popen.  Normally I think most users would want to know if a crash command output came back with non-text characters (garbage).  I think if we use binary we lose the ability to catch that exception right?

DaveWysochanskiRH: Ok I think I figured this out or am getting close.  I might still add an attempt to decode the output and throw an error to the log if we get unusual characters in the crash output but this way we should just dump the output to a file anyway and move on.  Having both the error and the output is best of both worlds.  Let me do some more testing.

DaveWysochanskiRH: @mkutlak  is commit 73eaded what you had in mind as far as changes here?
It is not fully done as I've run into some issues with bt_filter.
I've thought for a while these commands would be better in a default, post-retrace hook - the success or fail of a task probably should not depend on them.  However right now may be fine to just refactor and complete this patchset as a first step towards that goal.

mkutlak: > @mkutlak is commit [73eaded](https://github.com/abrt/retrace-server/commit/73eadeddc143377e548819d04358dce2be0a6204) what you had in mind as far as changes here?

Yeah, pretty much.

I would just use add a `mode` argument to `add_misc` and `set_backtrace`
```python3
def set_backtrace(self, backtrace, mode="w")

def add_misc(self, name, value, overwrite=False, mode="wb"):
```

DaveWysochanskiRH:
I think I've fixed all issues and these 3 patches are ready to merge.
f1bc200 Fix backtrace of bt_filter on Python3
9e79d41 Remove 'utf-8' encoding for run_crash_cmd and change related functions.
b9741d3 Refactor crash commands run after prepare_debuginfo into run_crash_cmdline helper

mkutlak:
Thank you @DaveWysochanskiRH

PR #246, opened 2019-07-29
Move run crash cmdline

DaveWysochanskiRH:


ernestask:
Could you update the description and commit messages with the rationale behind the changes?

DaveWysochanskiRH:
> Could you update the description and commit messages with the rationale behind the changes?

Honestly I'm not sure what you're asking for since the original commit messages explained the rationale and the 'why' in the paragraph.  For the first patch, do you want to know why I wanted to move from one class to the other?  For the second patch do not find the rationale of the common timeout and calling all crash commands from one place convincing?

DaveWysochanskiRH:
As far as location of the common function to run crash, an argument could be made that the 'run_crash_cmdline()' should be same location as the run_gdb() function which is inside retrace.py and outside of any class.  It's possible that is better and there should be another patch, or just merge these two into one and make that separate function.

Maybe another question is why do one function for any crash command?  We've seen some crash commands run a very long time or even in loops so that is one main driver to use one function.    Originally when I added ProcessCommunicateTimeout it was due to this but it was only used in one location for one crash command.  This is historical information in that patch though so I didn't include it.

ernestask:
I want to know if you’re fixing some specific issue or if this is just making things better in some other sense. Also helps with the “what” moments when going through git logs.

ernestask:
I also don’t see why historical information shouldn’t be included. Digging through the said history is very time-consuming.

DaveWysochanskiRH:
See improved commit messages and let me know if this is ok.

src/retrace/retrace.py
```diff
@@ -1862,41 +1870,32 @@ def prepare_debuginfo(self, vmcore, chroot=None, kernelver=None, crash_cmd=["cra
...
+        stdout, returncode = self.run_crash_cmdline(crash_normal, "mod\nquit")
+        if returncode == 1 and "el5" in kernelver.release:
             log_info("Unable to list modules but el5 detected, trying crash fixup for vmss files")
             crash_cmd.append("--machdep")
             crash_cmd.append("phys_base=0x200000")
             log_info("trying crash_cmd = " + str(crash_cmd))
-            child = Popen(crash_cmd + ["-s", vmcore, vmlinux], stdin=PIPE, stdout=PIPE, stderr=STDOUT,
-                          encoding='utf-8')
-            stdout = child.communicate("mod\nquit")[0]
+            # FIXME: mock
```
ernestask: “mock” what?

DaveWysochanskiRH: This was a notation that the code does not work if mock is used.  It is unrelated to this patch but a note to look at this in the future.

DaveWysochanskiRH: This was put in there for vmware vmcores which require a special commandline to load the crash tool (option "phys_base=0x200000").  We have other issues with vmware such as snapshot vmem + vmss files so it would be a good candidate for a patchset like "handle vmware memory images better".

DaveWysochanskiRH: I just checked our production system and this code is still used on a few vmcores so I probably should not just remove it right now unfortunately, though it could be slated for removal / rework in the future (say 12 mo).

ernestask: Okay, thanks for the explanation.

ernestask:
Sorry for being so anal about these things, but I prefer being verbose for the sake of my future self.

DaveWysochanskiRH:
Not a problem at all.  I can be as verbose as you want most likely!  I tend to err on that side but these patches were exception at the outset.  I think the added commit log was good to add.  Thanks for the review!

DaveWysochanskiRH:
FYI, I think this patchset broke module extraction from kernel-debuginfos.  I'm working on tracking it down now and hopefully will have a fix soon.

PR #247, opened 2019-08-02
Fix module extraction from kernel-debuginfo

DaveWysochanskiRH:
Commit 2199ab8 broke module extraction because it did handle the
byte arrary return from run_crash_cmdline properly.
Rather than a string, run_crash_cmdline returns a byte array.  When
we are obtaining the list of module object files to extract from crash's
'mod' command, these should be in the form of strings so we can pass
them to 'cpio'.  The reason for the default byte array return on
run_crash_cmdline is found in the discussion of 7ea86bd:
1) some crash commands (not the 'mod' command) may return non-'utf-8'
characters (for example a command that reads a lot of
memory from a vmcore that may be incomplete or damaged)
2) often we save the output of crash commands to a file so we do not
care necessarily if we have non-'utf-8' characters.

### DIFF
--- a/src/retrace/retrace.py
+++ b/src/retrace/retrace.py
@@ -1890,12 +1890,12 @@ class RetraceTask:
             return vmlinux
 
         modules = []
-        for line in stdout.splitlines():
+        for line in stdout.decode('utf-8').splitlines():
             # skip header
-            if b"NAME" in line:
+            if "NAME" in line:
                 continue
 
-            if b" " in line:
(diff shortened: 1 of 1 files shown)


DaveWysochanskiRH:
NOTE: It's not too great that we have a 'run_gdb' command that returns string and 'run_crash_cmdline' that returns a bytearray.  Until now we've not really put these two 'debuggers' under a common class so it may be ok but is not ideal.

PR #252, opened 2019-09-09
Remove extra crash commands

DaveWysochanskiRH:
This set of patches does the first portion of the following bug:
https://bugzilla.redhat.com/show_bug.cgi?id=1266769

It removes the extra crash commands from retrace-server, and enables a post-retrace 'success' hook to run them.  I'm confident the right thing to do is pull these out for reasons in the patch header.

As far as where they end up, I have a sample implementation for a 'success' hook that relies on pexpect library, which looks to be the very best option for post-analysis.  I will send in an internal email to you.  I am not entirely decided on:
1. Whether to rework existing hook scripts or try to keep it as it is (there are a number of issues)
2. Whether to try to ship some default 'hook' for 'success' script to provide these commands

For now I think the answer to both of the above is 'no' and 'no' but I'm less sure about #1 because of a couple issues.


mkutlak:
Thank you @DaveWysochanskiRH 

I will merge this and open a RFE for `retrace-server-hooks`.

PR #254, opened 2019-09-13
Clean up handling of mock with x86 vmcores in x86_64 environments

DaveWysochanskiRH:
Originally when RetraceTask.use_mock was introducted it was done
alongside the implementation of RetraceTask.crash_cmd.  When crash_cmd
was introduced it was not well understood where we really should set
this and read it.  Also, use_mock was not really well done either.

The right thing to do with crash_cmd and use_mock is only set both of
them once the arch has been properly detected from get_kernel_release.
Create methods get_mock(), set_mock(), and has_mock() to handle whether
mock is configured or not.

This cleans up the code quite a bit and fixes on bug which occurs
when 32-bit cores are run on a 64-bit machine where prepare_debuginfo
would be called twice:
https://bugzilla.redhat.com/show_bug.cgi?id=1536683

### DIFF
--- a/src/retrace-server-interact
+++ b/src/retrace-server-interact
@@ -111,31 +111,24 @@ if __name__ == "__main__":
             kernelver = get_kernel_release(vmcore, task.get_crash_cmd().split())
             if kernelver is None:
                 raise Exception("Unable to determine kernel version")
-            task.set_kernelver(str(kernelver))
+            task.set_kernelver(kernelver)
 
         hostarch = os.uname()[4]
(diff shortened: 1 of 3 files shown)


DaveWysochanskiRH:
This was a patch that was outstanding for a long time which fixes a long ago bug with 32-bit on 64-bit machine, and cleans up the handling of mock for vmcores.  I have been testing this on top of the other patches that were just merged as well as some hooks patches under test.

mkutlak:
LGTM.

PR #255, opened 2019-09-16
Pass results dir to hooks

DaveWysochanskiRH:
I've been testing these patches and I think they are a good modest change to the hooks interface.  If we stick with a lightweight interface to hooks and provide freedom for a hook script to interface with retrace-server via the existing retrace-server-* commands, the only thing I think we need to change is to give a hook a specific directory to write output (results files).  Note the renaming patch is backward compatible with tasks that have existing 'misc' subdirectory so this should not break anything.

I also thought of exporting the "add_results()" and similar functions to a hook script but I think it might be better to avoid that.


mkutlak:
LGTM

PR #257, opened 2019-09-24
Refactor into KernelVMcore and eliminate duplicate prepare_debuginfo

DaveWysochanskiRH:
This series of patches refactors a number of functions into the new KernelVMcore class, and eliminates some confusing and duplicate logic involving vmcore setup, especially the duplicate calls to prepare_debuginfo which happens on some vmcores.

There are likely further cleanups that can be done on this code but I felt this a fairly good series to move things forward.  I have tested this fairly thoroughly and I do not believe there are any regressions here, but it's always possible some corner case could have been missed.

### DIFF
--- a/src/create.wsgi
+++ b/src/create.wsgi
@@ -199,7 +199,8 @@ def application(environ, start_response):
                             _("Required file '%s' is missing") % required_file)
 
     if task.get_type() in [TASK_VMCORE, TASK_VMCORE_INTERACTIVE]:
-        task.strip_vmcore(os.path.join(crashdir, "vmcore"))
+        vmcore = KernelVMcore(os.path.join(crashdir, "vmcore"))
+        vmcore.strip_extra_pages(task)
 
     retcode = task.start()
     if retcode != 0:
--- a/src/retrace-server-interact
+++ b/src/retrace-server-interact
@@ -104,11 +104,13 @@ if __name__ == "__main__":
         sys.stderr.write("Action '%s' is not allowed for coredumps.\n" % args.action)
         exit(1)
     elif task.get_type() == TASK_VMCORE_INTERACTIVE:
-        vmcore = os.path.join(task.get_savedir(), "crash", "vmcore")
+        vmcore_path = os.path.join(task.get_savedir(), "crash", "vmcore")
+        vmcore = KernelVMcore(vmcore_path)
+
         if task.has_kernelver():
             kernelver = KernelVer(task.get_kernelver())
         else:
-            kernelver = get_kernel_release(vmcore, task.get_crash_cmd().split())
+            kernelver = vmcore.get_kernel_release(task.get_crash_cmd().split())
             if kernelver is None:
                 raise Exception("Unable to determine kernel version")
             task.set_kernelver(kernelver)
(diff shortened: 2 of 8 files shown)


src/retrace/retrace.py
```diff
@@ -1355,6 +1323,38 @@ def convert_flattened_format(self):
         self._is_flattened_format = False
         return True
 
+    DUMP_LEVEL_PARSER = re.compile(r"^[ \t]*dump_level[ \t]*:[ \t]*([0-9]+).*$")
```
mgrabovsky: It is customary to define member constants and variables at the top of the class (just after the `class Foo:` declaration). Can you please move it there?

src/retrace-server-interact
```diff
@@ -123,7 +123,20 @@ if __name__ == "__main__":
...
-                vmlinux = task.prepare_debuginfo(vmcore, kernelver=kernelver)
+                if self.has_status() and \
+                   self.get_status() not in [STATUS_SUCCESS, STATUS_FAIL]:
+                    sys.stderr.write("Task '%s' still in progress or hung (status = %d), "
+                                     "please wait for task to complete.\n" %
+                                     (task.get_taskid(), task.get_status()))
+                    sys.stderr.write("If you suspect task is hung and will never "
+                                     "complete, try forcing failure with "
+                                     "retrace-server-interact, and restart the task.")
+                    exit(1)
```
mgrabovsky: Use `sys.exit` instead, please. There are some tricky issues with plain `exit`.

DaveWysochanskiRH: do you want me to add a patch to replace exit() with sys.exit() for all retrace-* commands?

mgrabovsky: Even though it's slightly out of scope of the PR, it's still refactoring and it's a minor change, so I won't mind it. Go ahead.

mgrabovsky:
Seems all right. Thank you for the PR.

src/create.wsgi
```diff
@@ -199,7 +199,8 @@ def application(environ, start_response):
                             _("Required file '%s' is missing") % required_file)
 
     if task.get_type() in [TASK_VMCORE, TASK_VMCORE_INTERACTIVE]:
-        task.strip_vmcore(os.path.join(crashdir, "vmcore"))
+        vmcore = KernelVMcore(os.path.join(crashdir, "vmcore"))
+        vmcore.strip_extra_pages(task)
```
ernestask: I suppose this now needs `vmcore.prepare_debuginfo(task)` beforehand?

ernestask: And to drop the argument from the `strip_extra_pages()` call.

mgrabovsky: Indeed. Missed that in the review.

wysochanski: Nice catch!  I need a test for this. Do we have a tool that calls this method?  I used “retrace-server-task create” and the manager page.  But neither of these calls this create code.

ernestask: Good question, I thought `retrace-server-task create` would hit the path? It actually did so in my testing.

ernestask: I can look at the traces later to see if it reveals something more.

DaveWysochanskiRH: Ok I see it now.  There is 3 options for 'retrace-server-task create':
1. 'manager' - uses '__custom__' which is default and well tested, and references a file local to retrace-server.
2. 'ftp' - uses 'start' method.  This is non-default but is lightly used because it fetches the vmcore from ftp.
3. 'http' - this uses 'create' method.  This is non-default, and you have to send the vmcore over http.  It is not used much if at all by us so this explains why I missed it.  I'll need to review it and test.

A quick review and it looks like this definitely needs a closer look so I'll see what else may be needed.  Even with 7211122c4cf3e704cbab68a6767c7389150c0939 we are not handling flattened format for example via the 'create' code path.  I am not sure if I can merge the logic inside 'start_vmcore' with what is inside 'create' or if it truly needs duplicating.

PR #259, opened 2019-09-27
Update plugins

mkutlak:
The devtools-4 was EOL in Dec 2017.
softwarecollections.org/en/scls/rhscl/devtoolset-4

Current one:
softwarecollections.org/en/scls/rhscl/devtoolset-8
___
We have centos plugin in production but it is missing in upstream.
___
Updates to stats page to show latest 10 Fedora releases and RHEL8/CentOS8.


ernestask:
LGTM

PR #267, opened 2019-10-14
Typo

michalfabik:
See commit.


ernestask:
Double the fixing, double the funning.

PR #270, opened 2019-10-23
Minor refactoring of retrace-server

mkutlak:


src/create.wsgi
```diff
@@ -164,7 +164,7 @@ def application(environ, start_response):
 
         if unpack_retcode != 0:
             raise Exception
-    except:
+    except OSError:
```
ernestask: Is this right? The line just above raises an `Exception` exception.

mkutlak: :man_facepalming:

PR #273, opened 2019-10-30
Make retrace-server in docker functional

mkutlak:
For `less` see commits.


ernestask:
So what’s the deal with vim there?

mkutlak:
> So what’s the deal with vim there?

Muscle memory.

PR #276, opened 2019-11-01
retrace_worker: Remove distro hack from mock config

mkutlak:
Introduced in 5a1e21a0f6ca8174a88f4eaf37887100e0ec311b

As `dnf` is  now available on both RHEL(7,8) and Fedora (and with the `--skip-broken` switch support),
we can remove the `distro` "hack".

### DIFF
--- a/src/retrace/retrace_worker.py
+++ b/src/retrace/retrace_worker.py
@@ -9,7 +9,6 @@ import shutil
 import stat
 from subprocess import Popen, PIPE, STDOUT
 
-import distro
 from .retrace import (ALLOWED_FILES, INPUT_PACKAGE_PARSER, REPO_PREFIX, REQUIRED_FILES,
                       STATUS, STATUS_ANALYZE, STATUS_BACKTRACE, STATUS_CLEANUP,
                       STATUS_FAIL, STATUS_INIT, STATUS_STATS, STATUS_SUCCESS,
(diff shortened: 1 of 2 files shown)


ernestask:
It’s available, but nothing in the package dependency chain really pulls it in on RHEL 7, as far as I can tell.

mkutlak:
> It’s available, but nothing in the package dependency chain really pulls it in on RHEL 7, as far as I can tell.

True. 

Maybe I shouldn't mention the RHEL7 because upstream doesn't work on it. (missing Python3 dependencies)

src/retrace/retrace_worker.py
```diff
@@ -473,20 +472,14 @@ def start_retrace(self, custom_arch=None):
...
-                else:
-                    mockcfg.write("--skip-broken")
+                mockcfg.write("config_opts['chroot_setup_cmd'] = '--skip-broken")
                 mockcfg.write(" install %s abrt-addon-ccpp shadow-utils %s rpm'\n" % (" ".join(packages),
                                                                                       self.plugin.gdb_package))
-                mockcfg.write("config_opts['releasever'] = '%s'\n" % linux_dist[1])
-                if linux_dist[0] == "fedora":
-                    mockcfg.write("config_opts['package_manager'] = 'dnf'\n")
+                mockcfg.write("config_opts['releasever'] = '%s'\n" % version)
+                mockcfg.write("config_opts['package_manager'] = 'dnf'\n")
```
mkutlak: Should I remove this line then?

ernestask: I’m thinking not, since you say that we’re missing dependencies for actual support. Would be nice if things just worked, but oh well.

ernestask:
AAAAAAAAAAAAAAAAND MERGE

PR #278, opened 2019-11-22
Change of retrace hooks

mkutlak:
Instead of specifying what to run for certain a hook in `/etc/retrace-server.conf`
you will be able to put executable scripts into one of the hook directories in
`/usr/share/retrace-server/hooks/{start,fail,success,...}` and
retrace-server worker will run all the executable scripts it finds.

The directory can be changed with HookDir in `/etc/retrace-server/retrace-server-hooks.conf`.

For now you can configure `timeout` and `cmdline` options for the hooks.

Example for a `success` hook:
$ ls /usr/share/retrace-server/hooks/success/
/usr/share/retrace-server/hooks/success/echo

$ cat /etc/retrace-server/hooks/success.conf
```
[success]
timeout = 42

[success.echo]
cmdline = {taskid}
```
retrace-server worker will now run `echo 4321` with timeout of 42 seconds
once some retrace task finishes successfuly.

### DIFF
--- a/src/config/hooks/mock.conf
+++ b/src/config/hooks/mock.conf
@@ -1,0 +1,10 @@
+# Parameters are replaced using python's format.
+# Available parameters for cmdline: hook_name, taskid, task_results_dir
+# Example:
+# cmdline = {hook_name} {taskid} {task_results_dir}
+#
+# Before the preparation of retrace environment
+# [pre_prepare_mock]
+
+# After the preparation of retrace environment
+# [post_prepare_mock]
--- a/src/config/hooks/retrace.conf
+++ b/src/config/hooks/retrace.conf
@@ -1,0 +1,10 @@
+# Parameters are replaced using python's format.
+# Available parameters for cmdline: hook_name, taskid, task_results_dir
+# Example:
+# cmdline = {hook_name} {taskid} {task_results_dir}
+#
+# Before starting of the retracing itself
+# [pre_retrace]
+
+# After retracing is done
+# [post_retrace]
--- a/src/retrace/hooks/hooks.py
+++ b/src/retrace/hooks/hooks.py
@@ -1,0 +1,108 @@
+#!/usr/bin/python3
+
+import os
+import shlex
+from subprocess import PIPE, CalledProcessError, run, TimeoutExpired
+
+from retrace.retrace import log_info, log_error, log_debug
+from .config import HOOK_PATH, HOOK_TIMEOUT, hooks_config
+
+"""
+    Hooks description:
+    pre_start -- When self.start() is called
+    start -- When task type is determined and the main task starts
+    pre_prepare_debuginfo -- Before the preparation of debuginfo packages
+    post_prepare_debuginfo -- After the preparation of debuginfo packages
+    pre_prepare_mock -- Before the preparation of mock environment
+    post_prepare_mock -- After the preparation of mock environment
+    pre_retrace -- Before starting of the retracing itself
+    post_retrace -- After retracing is done
+    success -- After retracing success
+    fail -- After retracing fails
+    pre_remove_task -- Before removing task
+    post_remove_task -- After removing task
+    pre_clean_task -- Before cleaning task
+    post_clean_task -- After cleaning task
+"""
+
+
+def get_executables(path):
+    """ Scan `path` and return list of found executable scripts.
+    """
+    def _getname(entry):
+        return entry.name
+
+    script_list = []
+
+    if not os.path.isdir(path):
+        return script_list
+
+    with os.scandir(path) as dirls:
+        for entry in sorted(dirls, key=_getname):
+            if entry.is_file() and os.access(entry.path, os.X_OK):
+                script_list.append(entry)
+
+    return script_list
+
+
+class RetraceHook:
+
+    def __init__(self, task):
+        self.taskid = task.get_taskid()
+        self.task_results_dir = task.get_results_dir()
+
+    def _get_cmdline(self, hook, exc=None):
+        if exc:
+            cmdline = hooks_config.get(f"{hook}.{exc}.cmdline", None)
+
+        if not cmdline:
+            cmdline = hooks_config.get(f"{hook}.cmdline", None)
+
+        if cmdline:
+            cmdline = cmdline.format(hook_name=hook,
+                                     taskid=self.taskid,
+                                     task_results_dir=self.task_results_dir)
+
+        return cmdline
+
+    def _get_hookdir(self):
+        hooks_path = hooks_config.get("main.hookdir", HOOK_PATH)
+
+        return hooks_path
+
+    def _get_timeout(self, hook):
+        timeout = hooks_config.get("main.timeout", HOOK_TIMEOUT)
+
+        if f"{hook}.timeout" in hooks_config:
+            timeout = hooks_config.get(f"{hook}.timeout", timeout)
+
+        return int(timeout)
+
+    def run(self, hook):
+        """Called by the default hook implementations"""
+        hook_path = os.path.join(self._get_hookdir(), hook)
+        executables = get_executables(hook_path)
+        timeout = self._get_timeout(hook)
+
+        for exc in executables:
+            log_debug(f"Running '{hook}' hook - script '{exc.name}'")
+            script = exc.path
+            hook_cmdline = self._get_cmdline(hook, exc.name)
+
+            if hook_cmdline:
+                script = shlex.quote(f"{script} {hook_cmdline}")
+
+            script = shlex.split(script)
+            child = run(script, shell=True, timeout=timeout, stdout=PIPE, stderr=PIPE, encoding='utf-8')
+
+            try:
+                child.check_returncode()
+            except TimeoutExpired:
+                log_error(f"Hook script '{exc.name}' timed out ({timeout}s).")
+            except CalledProcessError:
+                log_error(f"Hook script failed with exit status {child.returncode}.")
+
+            if child.stdout:
+                log_info(child.stdout)
+            if child.stderr:
+                log_error(child.stderr)
--- a/src/retrace/hooks/hooks.py
+++ b/src/retrace/hooks/hooks.py
@@ -1,7 +1,10 @@
 #!/usr/bin/python3
 
+import itertools
+import multiprocessing as mp
 import os
 import shlex
+
 from subprocess import PIPE, CalledProcessError, run, TimeoutExpired
 
 from retrace.retrace import log_info, log_error, log_debug
@@ -40,7 +43,7 @@ def get_executables(path):
     with os.scandir(path) as dirls:
         for entry in sorted(dirls, key=_getname):
             if entry.is_file() and os.access(entry.path, os.X_OK):
-                script_list.append(entry)
+                script_list.append(entry.path)
 
     return script_list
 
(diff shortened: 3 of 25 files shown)


DaveWysochanskiRH:
Looks very nice!  I am going to take this into a testing machine and give it a test.  This improves on the out-of-tree directory executable extension I have right now so it will be nice to get rid of it and have this code in retrace-server.

DaveWysochanskiRH:
One problem I see with this is you still need root to install hook scripts, or am I misunderstanding?
I think any user with AuthGroup perms should be able to install a hook script, so maybe some perms are wrong on /usr/share/retrace-server/hooks/* directories?

$ touch /usr/share/retrace-server/hooks/success/sbr-analysis.py
touch: cannot touch '/usr/share/retrace-server/hooks/success/sbr-analysis.py': Permission denied
$ ls -ld /usr/share/retrace-server/hooks/success/
drwxr-xr-x 2 root root 6 Nov 22 09:37 /usr/share/retrace-server/hooks/success/

DaveWysochanskiRH:
Aside from being unable to install into the specific hook directory, I only made the following edit to the 'success.conf' file, copied the existing hook scripts over as root and they seem to run great!  I have not done extensive tests such as timeouts but looks like you have that in there.  I'll try to do some further testing later today and let you know what I find.  I think there is a problem with relative paths but I'll see if I can find out if there's a workaround or what $CWD is when executing the scripts.
$ grep -v ^# /etc/retrace-server/hooks/success.conf
[success]
cmdline = {taskid} {task_results_dir}

DaveWysochanskiRH:
Ok there is a problem with the working directory when a hook script runs.  By default $(cwd) looks like it is $HOME but shouldn't it be the location of the hook script being executed?

We can have both retrace-server executing tasks in which case $HOME will be one value and if a user executes 'retrace-server-worker --restart' $HOME will be that users home directory.  It would be good to have the hook script set $(cwd) to a consistent value regardless of user.

Example trivial 'success' script to demonstrate the problem:
```
# cat /usr/share/retrace-server/hooks/success/pwd.sh
#!/bin/bash
echo "PWD = $(pwd)"
```
retrace-log when run via retrace
```
    2019-11-22 13:15:08 Retrace job finished successfully
   2019-11-22 13:15:08 Running 'success' hook - script 'pwd.sh'
    2019-11-22 13:15:08 PWD = /usr/share/retrace-server
```

retrace-log when run as a user
```
    2019-11-22 13:14:07 Retrace job finished successfully
    2019-11-22 13:14:07 PWD = /home/dwysocha
```

mkutlak:
> One problem I see with this is you still need root to install hook scripts, or am I misunderstanding?
> I think any user with AuthGroup perms should be able to install a hook script, so maybe some perms are wrong on /usr/share/retrace-server/hooks/* directories?

I think the default hook directory should stay `/usr/share/retrace-server/hooks/` with root permissions. 

If want a different location for the hooks, you can change it in `/etc/retrace-server/retrace-server-hooks.conf` with `HookDir=` option.

> By default $(cwd) looks like it is $HOME but shouldn't it be the location of the hook script being executed?

I've changed the `cwd` for the hooks to task's save dir.

DaveWysochanskiRH:
> > One problem I see with this is you still need root to install hook scripts, or am I misunderstanding?
> > I think any user with AuthGroup perms should be able to install a hook script, so maybe some perms are wrong on /usr/share/retrace-server/hooks/* directories?
> 
> I think the default hook directory should stay `/usr/share/retrace-server/hooks/` with root permissions.
> 

Ok let me try the reconfiguration of the directories though it is another config step.

> If want a different location for the hooks, you can change it in `/etc/retrace-server/retrace-server-hooks.conf` with `HookDir=` option.
> 
> > By default $(cwd) looks like it is $HOME but shouldn't it be the location of the hook script being executed?
> 
> I've changed the `cwd` for the hooks to task's save dir.

The tasks' save dir is passed as a parameter so that's not necessary.  Please can you make CWD the directory where the hook is executing?  That way there is no issue with relative paths of hook scripts but each hook script will be properly oriented.

mkutlak:
> Please can you make CWD the directory where the hook is executing? That way there is no issue with relative paths of hook scripts but each hook script will be properly oriented.

Done.

Question about timeouts/exceptions in hooks: 
Should retracing be cancelled if a hook fails or time-outs or just log the error and continue with retracing?

DaveWysochanskiRH:
> > Please can you make CWD the directory where the hook is executing? That way there is no issue with relative paths of hook scripts but each hook script will be properly oriented.
> 
> Done.
> 
Thanks!

> Question about timeouts/exceptions in hooks:
> Should retracing be cancelled if a hook fails or time-outs or just log the error and continue with retracing?

I would just log the error and continue.

DaveWysochanskiRH:
> > > Please can you make CWD the directory where the hook is executing? That way there is no issue with relative paths of hook scripts but each hook script will be properly oriented.
> > 
> > 
> > Done.
> 
> Thanks!
> 
> > Question about timeouts/exceptions in hooks:
> > Should retracing be cancelled if a hook fails or time-outs or just log the error and continue with retracing?
> 
> I would just log the error and continue.

Actually this may depend on the type of hook but I'm not sure - for now I think log and continue.  Right now I don't have any hooks other than 'success' and my guess is we should not stop retrace for a hook since these are in essence "add on" to retrace-server.

DaveWysochanskiRH:
> > One problem I see with this is you still need root to install hook scripts, or am I misunderstanding?
> > I think any user with AuthGroup perms should be able to install a hook script, so maybe some perms are wrong on /usr/share/retrace-server/hooks/* directories?
> 
> ~I think the default hook directory should stay `/usr/share/retrace-server/hooks/` with root permissions.~
> EDIT:
> Is there any better place where to place the hooks?
> 
> If want a different location for the hooks, you can change it in `/etc/retrace-server/retrace-server-hooks.conf` with `HookDir=` option.
> 

Let me think about it.  What I had planned to do is have a cron based 'git pull' job to sync the hooks from a repo.  Different people have commit access to the repo and can add post-analysis scripts into 'success' hook area, but they won't necessarily change the files directly on the machine.

src/retrace/hooks/hooks.py
```diff
@@ -0,0 +1,108 @@
...
+
+        return int(timeout)
+
+    def run(self, hook):
+        """Called by the default hook implementations"""
+        hook_path = os.path.join(self._get_hookdir(), hook)
+        executables = get_executables(hook_path)
+        timeout = self._get_timeout(hook)
+
+        for exc in executables:
```
DaveWysochanskiRH: When I originally did this loop I had a similar sequential loop.  But I thought it would be better to start all executable scripts in parallel and then wait for the exits.  You may want to consider something like this:
``    pids = []
    for f in os.listdir(h):
        if os.path.isfile(h + "/" + f) and os.access(h + "/" + f, os.X_OK):
            logging.info("Executing hook script %s/%s - max runtime %d seconds)" % (h, f, HOOK_RUNTIME_MAX))
            os.chdir(h)
            p = subprocess.Popen(["./" + f, args['task_id'], args['task_results_dir']],
                                 stdout=subprocess.PIPE, stderr=subprocess.PIPE)
            pids.append(p)
            os.chdir(cwd)

    logging.info("Waiting for all hooks to exit")
    count = HOOK_RUNTIME_MAX
    while len(pids) and count:
        for p in pids:
            if p.poll() is not None:
                logging.info("Hook pid %d exited with %d while running %s" %\
                             (p.pid, p.returncode, p.args))
                (out, err) = p.communicate()
                logging.info(out.decode("utf-8", "ignore"))
                logging.error(err.decode("utf-8", "ignore"))
                pids.remove(p)
        time.sleep(1)
        count -= 1

    if len(pids):
        for p in pids:
            p.kill()
            logging.warn("Hook pid %d got killed - %d second max runtime exceeded while running %s" %\
                         (p.pid, HOOK_RUNTIME_MAX, p.args))

``

DaveWysochanskiRH:
> > > One problem I see with this is you still need root to install hook scripts, or am I misunderstanding?
> > > I think any user with AuthGroup perms should be able to install a hook script, so maybe some perms are wrong on /usr/share/retrace-server/hooks/* directories?
> > 
> > 
> > ~I think the default hook directory should stay `/usr/share/retrace-server/hooks/` with root permissions.~
> > EDIT:
> > Is there any better place where to place the hooks?

FWIW, I tested the configurable location and it works fine for my use case.

As far as default, I'm not sure if /usr/share is a good place since that is for "architecture independent" files:
http://www.pathname.com/fhs/pub/fhs-2.3.html#USRSHAREARCHITECTUREINDEPENDENTDATA

Maybe /srv/retrace-server/hooks ?
http://www.pathname.com/fhs/pub/fhs-2.3.html#SRVDATAFORSERVICESPROVIDEDBYSYSTEM
"This main purpose of specifying this is so that users may find the location of the data files for particular service, and so that services which require a single tree for readonly data, writable data and scripts (such as cgi scripts) can be reasonably placed. "

As you know, there are other configurable directories in retrace-server.conf:
# Log directory
LogDir = /var/log/retrace-server

# Local repos directory
# if changed, you also need to update httpd config
RepoDir = /var/cache/retrace-server

# Directory where the crashes and results are saved
SaveDir = /var/spool/retrace-server

# Directory where old tasks are moved
DropDir = /srv/retrace/archive

 

> > If want a different location for the hooks, you can change it in `/etc/retrace-server/retrace-server-hooks.conf` with `HookDir=` option.
> 
> Let me think about it. What I had planned to do is have a cron based 'git pull' job to sync the hooks from a repo. Different people have commit access to the repo and can add post-analysis scripts into 'success' hook area, but they won't necessarily change the files directly on the machine.

mkutlak:
> As far as default, I'm not sure if /usr/share is a good place since that is for "architecture independent" files:
> http://www.pathname.com/fhs/pub/fhs-2.3.html#USRSHAREARCHITECTUREINDEPENDENTDATA
> 
> Maybe /srv/retrace-server/hooks ?


I moved the default location to `/usr/libexec/retrace-server/hooks` and `retrace` group has `rwx` permissions now.

Also you can now as a user change the `HookDir` location (or any other configuration) in `$HOMEDIR/.config/retrace-server/retrace-server-hooks.conf` config.


```
The configs are read and evaluated in this order:

1) /etc/retrace-server/hooks/*.conf
2) /etc/retrace-server/retrace-server-hooks.conf
3) $HOME/.config/retrace-server/hooks/*.conf
4) $HOME/.config/retrace-server/retrace-server-hooks.conf
```

src/retrace/hooks/hooks.py
mkutlak: Done.

DaveWysochanskiRH:
I tested these patches including commit 85d2fba including a few different timeout settings.
I appreciate adding the parallel execution of all executables as I think that will speed up post-processing and be quite useful.  These patches look good to me and seem to work as advertised for major functions.

For commit 5e2fc0e, I'm not sure did you intend for the user-based config to override the default settings?  Also, what is the use-case for having this?  It seems like it will grab one set of configs if a core is queued from the manager interface, and a different one if done from the commandline (retrace-server-task create).  Is that desired?  I could not get this to work though I didn't extensively debug it.  I tried copying the retrace-server-hooks.conf file and the 'success.conf' file into ~/.config/retrace-server and ~/.config/retrace-server/hooks/ respectively.

DaveWysochanskiRH:
Oh there is one more point.  At some point it _might_ be useful to have per-executable timeouts, as you have configurable per-executable commandlines.  However for now I would say that is probably overkill so unless there's a valid need, it's not necessary.

mkutlak:
> per-executable timeouts

Done.

DaveWysochanskiRH:
Extra credit!  Great - I will test this out and try to do some more testing of the configs as well.

src/config/hooks/mock.conf
```diff
@@ -0,0 +1,10 @@
+# Parameters are replaced using python's format.
+# Available parameters for cmdline: hook_name, taskid, task_results_dir
+# Example:
+# cmdline = {hook_name} {taskid} {task_results_dir}
+#
+# Before the preparation of mock environment
```
michalfabik: suggest the more general description "retrace environment" because it can be either mock or podman

src/config/hooks/retrace.conf
```diff
@@ -0,0 +1,10 @@
+# Parameters are replaced using python's format.
+# Available parameters for cmdline: hook_name, taskid, task_results_dir
+# Example:
+# cmdline = {hook_name} {taskid} {task_results_dir}
+#
+# Before starting of the retracing itself
+# [pre_retrace_retrace]
```
michalfabik: `pre_retrace_retrace` and `post_retrace_retrace` not consistent with the description at src/retrace/hooks/hooks.py:21,22

mkutlak: "Typo", fixed.

src/config/hooks/mock.conf
mkutlak: Changed.

michalfabik: And the config item itself?

mkutlak: Changed with last commit.

DaveWysochanskiRH:
I'm getting a build error after this merge due to last commit 83327af so might need a fixup?
Builds ok if I "git reset --hard 9cdd174"
Machine is f30 and the following build line:  ./autogen.sh; ./configure; make rpm

...
make[7]: Leaving directory '/root/git/retrace-server/src/config'
make[6]: *** [Makefile:607: distdir] Error 2
make[6]: Leaving directory '/root/git/retrace-server/src/config'
make[5]: *** [Makefile:731: distdir-am] Error 1
rm retrace-server-cleanup.xml retrace-server-task.xml retrace-server-interact.xml retrace-server-worker.xml retrace-server-reposync.xml
make[5]: Leaving directory '/root/git/retrace-server/src'
make[4]: *** [Makefile:727: distdir] Error 2
make[4]: Leaving directory '/root/git/retrace-server/src'
make[3]: *** [Makefile:579: distdir-am] Error 1
make[3]: Leaving directory '/root/git/retrace-server'
make[2]: *** [Makefile:573: distdir] Error 2
make[2]: Leaving directory '/root/git/retrace-server'
make[1]: *** [Makefile:675: dist] Error 2
make[1]: Leaving directory '/root/git/retrace-server'
make: *** [Makefile:892: rpm] Error 2

DaveWysochanskiRH:
Looks like there was one other issue with this missed.  The existing /etc/retrace-server.conf file is not copied over to /etc/retrace-server/retrace-server.conf so you'll need to move it manually.  I have not tracked this down yet but ran into it after an upgrade.  It's not a huge deal though since it should be a one time move, and the old file is preserved (I think at /etc/retrace-server.conf.rpmsave)

mkutlak:
I had solution for this but it wasn't 100% reliable with updates/downgrades and I dropped it after discussion with others.

It is an annoying manual work, but as you say it should be just a one time thing.

I will document it in DEPLOYING wiki together with more information about the hooks in docs.

DaveWysochanskiRH:
Sounds good thanks!  Good to know you knew about it as I started wondering about following updates but figured those would not be affected.

PR #295, opened 2020-02-04
GPG keys changes

mkutlak:
Changes:
- Dropped the ancient GPG keys for Fedora 15,16,17.
- Enabled GPG check for rawhide.
- Use GPG keys from `distribution-gpg-keys`.
- As a side-effect you can add more GPG keys to distribution plugins (i.e. for packages from [Copr](https://copr.fedorainfracloud.org/) repositories)


src/retrace/retrace_worker.py
```diff
@@ -432,8 +429,7 @@ def start_retrace(self, custom_arch=None):
                     mockcfg.write("name=%s\n" % releaseid)
                     mockcfg.write("baseurl=file://%s/\n" % repopath)
                     mockcfg.write("failovermethod=priority\n")
-                    if version != "rawhide" and CONFIG["RequireGPGCheck"]:
-                        mockcfg.write("gpgkey=file:///usr/share/retrace-server/gpg/%s-%s\n" % (distribution, version))
+                    mockcfg.write("gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-fedora-%s-%s" % (version, arch))
```
xsuchy: Hmmm, does EL8 knows about fedora keys? Can we use distribution-gpg-keys package?

xsuchy:
Otherwise +1 to the removal of old keys.

src/retrace/retrace_worker.py
mkutlak: Bit of a chicken-egg situation because we need to install `distribution-gpg-keys` first, for which we need a GPG key from `distribution-gpg-keys`?

xsuchy: You can install it on host.

mkutlak:
More shenanignas will be needed for `rawhide` because rawhide needs more(?) keys (of older releases) to verify the packages.

mkutlak:
Fedora works with `mock` and I will test it with `podman` once `fedora:rawhide` is 33.

mkutlak:
Tested locally for `Fedora rawhide` (work for both `mock` and `podman`) and for
`CentOS 8` (works on `podman`, but there are some problems with `mock` and modular packages)

PR #299, opened 2020-02-17
Update regexes for CentOS and RHEL releases

mkutlak:
The release version is no longer just a simple number.

CentOS:
`CentOS Linux release 7 (Core)`
`CentOS Linux release 8.1.1234 (Core)`

RHEL:
`Red Hat Enterprise Linux release 8.1 (Ootpa)`
`Red Hat Enterprise Linux release 8.2 Beta (Ootpa)`
`Red Hat Enterprise Linux Server release 7.8 Beta (Maipo)`

https://regex101.com/

### DIFF
--- a/src/plugins/centos.py
+++ b/src/plugins/centos.py
@@ -2,7 +2,7 @@ import re
 
 distribution = "centos"
 abrtparser = re.compile(r"^CentOS Linux release (\d+)(?:\.(\d+)\.(\d+))? \(([^\)]+)\)$")
-guessparser = re.compile(r"\.el([0-9]+)")
+guessparser = re.compile(r"\.el(\d+)")
 displayrelease = "CentOS release"
 gdb_package = "gdb"
 gdb_executable = "/usr/bin/gdb"
--- a/src/plugins/centos.py
+++ b/src/plugins/centos.py
@@ -1,7 +1,7 @@
 import re
 
 distribution = "centos"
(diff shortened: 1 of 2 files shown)


src/plugins/centos.py
```diff
@@ -1,7 +1,7 @@
 import re
 
 distribution = "centos"
-abrtparser = re.compile(r"^CentOS Linux release ([0-9]+) \(([^\)]+)\)$")
+abrtparser = re.compile(r"^CentOS Linux release (\d+)(?:\S(\d+)\S(\d+))? \(([^\)]+)\)$")
 guessparser = re.compile(r"\.el([0-9]+)")
```
michalfabik: How about using `\d` on this line too, instead of `[0-9]`?

mkutlak: Sure.

src/plugins/centos.py
```diff
@@ -1,7 +1,7 @@
 import re
 
 distribution = "centos"
-abrtparser = re.compile(r"^CentOS Linux release ([0-9]+) \(([^\)]+)\)$")
+abrtparser = re.compile(r"^CentOS Linux release (\d+)(?:\S(\d+)\S(\d+))? \(([^\)]+)\)$")
```
michalfabik: Would it make sense to use explicit `\.` instead of `\S`?

mgrabovsky:
Looks good :+1:

PR #306, opened 2020-03-13
Fix alerts found by LGTM

mkutlak:
See commits for more.

- Removed never used variables
- Removed redundant else statement
- Fixed implicit concatenation
- Fixed too few parameters for `__setattr__` method
- Removed unnecessary assignment


mkutlak:
:pray:

mgrabovsky:
Nice! Looks good to me.

PR #309, opened 2020-03-31
Fix make check

michalfabik:
This PR fixes several minor errors in `test/` that caused `make check` to fail.


ernestask:
Please fill in the commit messages, so anyone reviewing has clear context of what’s going on.

ernestask:
>Some recent changes to the code seem to have introduced errors to run_test.py which apparently went unnoticed since make check isn't routinely run as a part of the build process.

That doesn’t really explain which changes caused what errors and why your code fixes them (https://github.com/torvalds/subsurface-for-dirk/blob/a48494d2fbed58c751e9b7e8fbff88582f9b2d02/README#L88 has some good advice).

test/run_test.py
```diff
@@ -232,7 +233,7 @@ def create_local_config(change_repo_dir):
 
 # instead of task.start() (a few missing lines, should be added?)
 cmdline = ["retrace-server-worker", "%d" % task.get_taskid(), "--foreground"]
-call(cmdline)
+run(cmdline, stdout=PIPE, stderr=PIPE, encoding='utf-8')
```
mkutlak, line 236: Why `stdout/stderr/encoding` when there is nothing done with the output afterwards?

PR #311, opened 2020-04-03
Add '-y' to autogen.sh option to install dependencies without prompting

DaveWysochanskiRH:
When a user passes 'sysdeps --install' to autogen.sh this will stop
with a prompt today and there's no way to run unattended.


ernestask:
I think you could pipe `yes` to it, but sure.

PR #315, opened 2020-04-05
Add support for Meson (w\ Tito and Packit)

ernestask:
This PR achieves the following:
  - Adds support for Meson, to enable faster builds with better maintainability.
  - Adds Tito configuration, for easier release management and package builds.
  - Adds Packit configuration, for CI purposes.

Autotools support is not nuked yet, but will be soon.


.tito/custom/abrt/tagger/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+        version = version.split('-', maxsplit=1)[0]
+
+        run_command('meson rewrite kwargs set project / version %s' % (version))
+        run_command('git add -- meson.build')
+
+    def _tag_release(self):
+        new_version = self._bump_version()
+        self._check_tag_does_not_exist(self._get_new_tag(new_version))
+        self._set_meson_project_version(new_version)
+        self._update_package_metadata(new_version)
```
xsuchy, line 15: Can you contribute this to Tito upstream?

ernestask: I can look into adding the Meson bits, but that’s kind of neither here nor there. I’m still going to need a custom tagger to get rid of the unneeded bloat.

ernestask:
@DaveWysochanskiRH, this might also affect you down the line. I’m not quite sure what systems you use retrace-server on in general, but RHEL 8 ships a version of Meson that’s not terribly out of date.

PR #310, opened 2020-04-03
Replace os.path functions with their pathlib equivalents

michalfabik:
This PR replaces (mostly) the `path` functions from the `os` module with their equivalents from `pathlib` in a (rather feeble) attempt to make the code more readable, especially awkward constructs lilke 'os.path.join`.  
  
This is probably a lot less useful than I thought originally, so feel free to suggest complete rewrites or just dumping this completely. I figure it might potentially have some sort of an impact in the future if/when stricter type checking is introduced with the [typing module](https://docs.python.org/3/library/typing.html).   
  
Thorough testing is still needed.

PR #324, opened 2020-06-02
Fix misuse of pathlib.Path

michalfabik:
See commit.  

Fix error in already merged https://github.com/abrt/retrace-server/pull/310


ernestask:
I already have a commit in my PR for reposync.

Issue #262, opened 2019-10-05
[python-requests] Header values must be strings


ernestask:
Since python-requests 2.11.0, header values must be strings: https://github.com/psf/requests/commit/2669ab797ce769ecedf5493b04cb976f33e37210 (and later https://github.com/psf/requests/commit/be31a90906deb5553c2e703fb05cf6964ee23ed5)

https://github.com/abrt/retrace-server/blob/7211122c4cf3e704cbab68a6767c7389150c0939/src/retrace-server-task#L159-L160

ernestask:
Observable in retrace-server 1.19.0:

```
Traceback (most recent call last):
  File "/usr/bin/retrace-server-task", line 624, in <module>
    create_new_task(ARGS)
  File "/usr/bin/retrace-server-task", line 169, in create_new_task
    prepped = req.prepare()
  File "/usr/lib/python3.6/site-packages/requests/models.py", line 267, in prepare
    hooks=self.hooks,
  File "/usr/lib/python3.6/site-packages/requests/models.py", line 314, in prepare
    self.prepare_headers(headers)
  File "/usr/lib/python3.6/site-packages/requests/models.py", line 448, in prepare_headers
    check_header_validity(header)
  File "/usr/lib/python3.6/site-packages/requests/utils.py", line 943, in check_header_validity
    "bytes, not %s" % (name, value, type(value)))
requests.exceptions.InvalidHeader: Value for header {Content-Length: 23221178} must be of type str or bytes, not <class 'int'>
```

PR #325, opened 2020-06-08
retrace-server-task: Stringify HTTP headers

ernestask:
python-requests strongly prefers header values to be strings and nothing
else these days.

Fixes https://github.com/abrt/retrace-server/issues/262


ernestask:
Seems that this is the only affected code. Could have fixed it right there and then.

Issue #329, opened 2020-06-09
retrace-server-task create fails with "Crash directory does not contain required file" even though file exists and is readable


DaveWysochanskiRH:
After b5a2bda4725b retrace-server-task create fails with:

    [2020-06-09 08:07:53] [I] Post-processing downloaded file
    [2020-06-09 08:07:53] [I] Analyzing crash data
    [2020-06-09 08:07:53] [E] Crash directory does not contain required file 'vmcore'

ernestask:
Spoiler:
https://github.com/abrt/retrace-server/blob/b5a2bda4725bf5a002d0077d22b5e0f06faec608/src/retrace/retrace.py#L1695

There’s more breakage after this.

ernestask:
https://github.com/abrt/retrace-server/blob/b5a2bda4725bf5a002d0077d22b5e0f06faec608/src/retrace/retrace.py#L1128

ernestask:
https://github.com/abrt/retrace-server/blob/b5a2bda4725bf5a002d0077d22b5e0f06faec608/src/retrace/retrace.py#L1134

ernestask:
If I may repeat myself: https://github.com/abrt/retrace-server/pull/310#issuecomment-608381441

All this either grows tests, gets fixed or gets reverted.

PR #331, opened 2020-06-09
Fix '+' operator use with Paths

michalfabik:
Leftover from https://github.com/abrt/retrace-server/pull/310

Resolves:
https://github.com/abrt/retrace-server/issues/329
Signed-off-by: Michal Fabik <mfabik@redhat.com>

### DIFF
--- a/src/retrace/retrace.py
+++ b/src/retrace/retrace.py
@@ -1125,13 +1125,13 @@ class KernelVMcore:
                 kernel_path = kernel_path + "."
             kernel_path = kernel_path + str(kernelver.flavour)
 
-        vmlinux_cache_path = debugdir_base + "usr/lib/debug/lib/modules" / kernel_path / "vmlinux"
+        vmlinux_cache_path = debugdir_base / "usr/lib/debug/lib/modules" / kernel_path / "vmlinux"
         if vmlinux_cache_path.is_file():
-            log_info("Found cached vmlinux at path: " + vmlinux_cache_path)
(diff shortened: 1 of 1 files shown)


src/retrace/retrace.py
```diff
@@ -1125,13 +1125,13 @@ def prepare_debuginfo(self, task, chroot=None, kernelver=None):
                 kernel_path = kernel_path + "."
             kernel_path = kernel_path + str(kernelver.flavour)
 
-        vmlinux_cache_path = debugdir_base + "usr/lib/debug/lib/modules" / kernel_path / "vmlinux"
+        vmlinux_cache_path = debugdir_base / "usr/lib/debug/lib/modules" / kernel_path / "vmlinux"
         if vmlinux_cache_path.is_file():
-            log_info("Found cached vmlinux at path: " + vmlinux_cache_path)
+            log_info("Found cached vmlinux at path: " + str(vmlinux_cache_path))
```
xsuchy, line 1130: I would slightly prefer:

`"Found cached vmlinux at path: {}".format(vmlinux_cache_path)

Many people find it more readable.
`

michalfabik: Thanks! It's actually safer because `.format()` works even if the arg is a `Path`, unlike `+`.

ernestask:
Good to go or should I wait?

michalfabik:
Good to go. \*fingers crossed\*

PR #337, opened 2020-06-10
More type annotations and fixing errors

mgrabovsky:
I started doodling around with type annotations and mypy and accidentally fixed a few more type errors.

### DIFF
--- a/src/retrace/hooks/config.py
+++ b/src/retrace/hooks/config.py
@@ -34,13 +34,13 @@ def load_hook_config() -> Dict[str, str]:
     hook_configs += get_config_files(MAIN_HOOK_CONFIGS_PATH)
 
     main_hook_config_file = Path(MAIN_CONFIG_PATH, MAIN_HOOK_CONFIG_FILE)
-    hook_configs.append(str(main_hook_config_file))
+    hook_configs.append(main_hook_config_file)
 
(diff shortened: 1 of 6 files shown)


mgrabovsky:
Found and fixed a few more, although there are still many remaining (and a couple false positives that I don't feel like dealing with right now).

It appears some of these changes were not really worth it to begin with. For instance, `repopath` defined as `Path` [in `retrace_worker.py`](https://github.com/abrt/retrace-server/pull/337/commits/0179547f0cd0c80e8b3e0056a2612376f3f6a8fb#diff-15145150e9d46bc5630e27606be82388R427) is only used as string later, `os.path.join()` would be perfectly fine here.

michalfabik:
Looks all good to me.  
   
> It appears some of these changes were not really worth it to begin with.  
  
Well, yeah. That dawned on me pretty soon after I started messing with pathlib. My initial motivation was to use this as a sort of gentle introduction into the type annotation/checking so I figured at least there's this one purpose it might serve. Now with all the breakage, I suppose the type checks are pretty much a necessity. Oh well. Sorry guys.

mgrabovsky:
Well, it's not a huge deal, but I suppose it's a bit annoying for our users. I, for one, am quite excited about type annotations, so I'm glad they're proliferating.

mgrabovsky:
Rebased. Ready for review.

PR #339, opened 2020-06-11
Add GitHub Actions workflow for running code checks

ernestask:
This I would call “going all-in” as far as type annotations go.

Feel free to argue about the configuration.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,0 +1,23 @@
+[mypy]
+disallow_untyped_calls = True
+pretty = True
+show_column_numbers = True
+show_error_codes = True
+show_error_context = True
+strict_equality = True
+warn_unreachable = True
+
+[mypy-bugzilla]
+ignore_missing_imports = True
+
+[mypy-dnf.*]
+ignore_missing_imports = True
+
+[mypy-hawkey]
+ignore_missing_imports = True
+
+[mypy-magic]
+ignore_missing_imports = True
+
+[mypy-sqlalchemy.*]
+ignore_missing_imports = True
--- a/retrace-server.spec
+++ b/retrace-server.spec
@@ -15,11 +15,15 @@ BuildArch: noarch
 
 BuildRequires: asciidoc
 BuildRequires: gettext
+BuildRequires: gzip
+BuildRequires: lsof
 BuildRequires: lsof
 BuildRequires: meson
 BuildRequires: python3-devel
+BuildRequires: tar
 BuildRequires: texinfo
 BuildRequires: xmlto
+BuildRequires: xz
 
 Requires: rsync
 Requires: mock >= 1.4.7
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -1,0 +1,78 @@
+name: Static analysis
+on:
+  pull_request:
+    branches: [master]
+  push:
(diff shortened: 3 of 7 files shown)


mypy.ini
```diff
@@ -0,0 +1,22 @@
...
+ignore_missing_imports = True
+
+[mypy-hawkey]
+ignore_missing_imports = True
+
+[mypy-magic]
+ignore_missing_imports = True
+
+[mypy-pyfaf.*]
+ignore_missing_imports = True
```
mgrabovsky, line 22: We could perhaps add a reminder to remove this once pyfaf gets type annotations on its own.

mypy.ini
```diff
@@ -0,0 +1,22 @@
+[mypy]
+disallow_untyped_defs = True
```
mgrabovsky, line 2: So this will keep screaming at us until all the functions are annotated. For someone reading the report, there's a lot of noise and it's hard to find "genuine" type errors among all the "this functions is not annotated".

[`disallow_untyped_calls`](https://mypy.readthedocs.io/en/stable/config_file.html#untyped-definitions-and-calls) might be a milder, interim alternative, so we could annotate bottom-up (from the perspective of call trees) without getting outright errors.

I think I'd slightly prefer the latter, but I'm not adamant.

mypy.ini
```diff
@@ -0,0 +1,22 @@
...
+pretty = True
+show_column_numbers = True
+show_error_codes = True
+show_error_context = True
+strict_equality = True
+
```
mgrabovsky, line 8: [`warn_unreachable`](https://mypy.readthedocs.io/en/stable/config_file.html#configuring-warnings) also looks as a potentially useful flag.

retrace-server.spec
```diff
@@ -20,6 +20,10 @@ BuildRequires: meson
 BuildRequires: python3-devel
 BuildRequires: texinfo
 BuildRequires: xmlto
+BuildRequires: gzip
+BuildRequires: lsof
+BuildRequires: tar
+BuildRequires: xz
```
mgrabovsky, line 26: Just my OCD: Can you sort all the `BuildRequires` alphabetically?

mypy.ini
ernestask: I can go either way, this was just to incentivize us to get it all done.

mypy.ini
ernestask: Sounds good.

mypy.ini
ernestask: I think this is more because FAF is not installed in the build environment, but I see your point. I could do it just for the mypy job.

retrace-server.spec
ernestask: Sure, no problem.

mypy.ini
mgrabovsky: Ah, OK, let's go all in.

mypy.ini
ernestask: All right, so now I install faf and generate the stubs, but the resulting number of errors is way too big, so I ended up using `disallow_untyped_calls` instead.

ernestask: Also mind that some of the stubs are incomplete, because of `__all__` not listing out everything.

PR #341, opened 2020-06-16
Clean up logic for mock and vmcores and add configurable debugger path option

DaveWysochanskiRH:
These patches remove the old set_mock() / get_mock() logic that was used to allow 32-bit crash to run on a 64-bit machine.  This is not used anymore and was fairly ugly code.

In addition, this adds a "native" mode to "RetraceEnvironment" config variable which is the ability to run on the native machine, not in mock or podman.  This mode is used for interactive mode of retrace-server for vmcores, but is not used for userspace coredumps.

Finally, we add a "KernelDebuggerPath" config option.  This allows us to set the path to crash to be some non-system path which we can build and deploy to, and cleans up the hardcoded path.

### DIFF
--- a/src/config/retrace-server.conf
+++ b/src/config/retrace-server.conf
@@ -133,8 +133,9 @@ UseFafPackages = 0
 # Spool directory for FAF packages
 FafLinkDir = /var/spool/faf
 
-# Run the retrace in a Mock chroot (default), or a Podman container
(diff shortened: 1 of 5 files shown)


DaveWysochanskiRH:
I have resolved the conflict in ba3d70d  and re-pushed.

PR #345, opened 2020-06-29
retrace: Fix error when restarting a task with retrace-server-worker

DaveWysochanskiRH:
Before this patch, retrace-server-worker --restart command fails with
[E] 'NoneType' object has no attribute 'split'

Fix the above by making sure an exising task has a 'crash_cmd' file
and if not, write a default one.


ernestask:
> Only users with write or admin permissions to the repository can trigger Packit-as-a-Service

Fixed.

Issue #346, opened 2020-06-29
Users unable run retrace-server-worker --restart due to write permissions to /cores/retrace/repos/ 


DaveWysochanskiRH:
We're had a problem for a while but it would be good to solve it.  Basically if a user has a vmcore with a custom kernel-debuginfo and they either don't know this or they forget and submit a vmcore to retrace-server the task will fail.  Then they copy the kernel-debuginfo into /cores/retrace/repos/download and use: retrace-server-worker --restart <taskid>.  This command fails though because the user cannot create a directory inside /cores/retrace/repos/kernel to extract the debuginfo and an error like this goes into the log:
~~~
[2020-06-28 19:59:43] [I] Searching for kernel-debuginfo package for 3.10.0-1127.el7.bz1400159.dbg5.x86_64
[2020-06-28 20:00:38] [E] prepare_debuginfo failed: Failed vmlinux caching from debuginfo at location: /cores/retrace/repos/kernel/x86_64/usr/lib/debug/lib/modules/3.10.0-1127.el7.bz1400159.dbg5.x86_64/vmlinux
~~~

After this point, there is one more error relating to the 'stats.db' file, in the _fail() method.  The stats.db file too is only writeable by retrace user so this fails:
~~~
[2020-06-28 20:00:38] [W] Failed to save crash statistics: attempt to write a readonly database
~~~

I am not sure exactly what approach to use to fix these problems.  If there was some way to submit these commands so that some 'retrace' owner task would pick them up (i.e. through httpd I guess) then these issues would disappear.  Failing that we could try some permissions tricks on the files / directories but I'm not sure that is a good approach.

DaveWysochanskiRH:
Forgot to say - right now you'll only see this problem if you first apply https://github.com/abrt/retrace-server/pull/345

ernestask:
We kinda sorta had a similar problem in ABRT itself and what we ended up doing was setting setgid on the binary and ensuring that the directories are group-writable.

DaveWysochanskiRH:
Actually there are more problems with "retrace-server-worker --restart" than I thought.  I just saw another error:
~~~
[2020-06-29 12:58:14] [E] [Errno 13] Permission denied: '/cores/retrace/tasks/974590112/md5sum'
~~~
And there's no group write perms on that, understandably:
~~~
$ ls -lh /cores/retrace/tasks/974590112/md5sum
[...]

PR #348, opened 2020-07-04
retrace: Add restart to retrace-server-task to restart an existing task

DaveWysochanskiRH:
It makes more sense to use retrace-server-task to restart a task the
same way we create a task.  In this way, we avoid numerous problems
with user credentials such as storing task files with user ownership
as well as issues with extracting kernel-debuginfos.

Fixes https://github.com/abrt/retrace-server/issues/346

### DIFF
--- a/man/retrace-server-task.txt
+++ b/man/retrace-server-task.txt
@@ -13,6 +13,10 @@ SYNOPSIS
                              [-e EMAIL] [-c CASENO] [-bz BUGZILLANO]
                              COREFILE
 
+'retrace-server-task' restart [-h] [-s SERVER] [-n] [-v]
+                              [-r KERNELVER]
+                              TASK_ID
+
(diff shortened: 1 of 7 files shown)


DaveWysochanskiRH:
This patch works to restart a task through 'retrace-server-task restart' and via the manager interface.  It actually was not bad to do at all.

However, there is a bit of cleanup still to do:
- various parameters passed in POST
- error return from request
- disable 'retrace-server-worker --restart' for non-retrace user (redirect to 'retrace-server-task restart')
- update man pages and help
- ?probably other things I missed

DaveWysochanskiRH:
I think this is a decent set of patches and it will fix #346.  I chose to keep the set of options minimal for now and mirrors 'retrace-server-worker --restart', though we could add other options maybe in another set of patches if needed.

PR #349, opened 2020-07-10
retrace: Fix class declaration order

michalfabik:
There were type check errors because some functions were annotated
with types/classes declared later in the code. This commit reorders
the classes as appropriate.

Signed-off-by: Michal Fabik <mfabik@redhat.com>

### DIFF
--- a/src/retrace/retrace.py
+++ b/src/retrace/retrace.py
@@ -391,89 +391,6 @@ def is_package_known(package_nvr: str, arch: str, releaseid: Optional[str] = Non
     return any([f.is_file() for f in candidates])
 
 
-def find_kernel_debuginfo(kernelver: KernelVer) -> Optional[Path]:
-    vers = [kernelver]
-
-    for canon_arch, derived_archs in ARCH_MAP.items():
-        if kernelver.arch == canon_arch:
-            vers = []
-            for arch in derived_archs:
-                cand = KernelVer(str(kernelver))
-                cand.arch = arch
-                vers.append(cand)
-
-    if CONFIG["UseFafPackages"]:
-        from pyfaf.storage import getDatabase
-        from pyfaf.queries import get_package_by_nevra
-        db = getDatabase()
-        for ver in vers:
-            p = get_package_by_nevra(db, ver.package_name_base(debug=True),
-                                     0, ver.version, ver.release, ver._arch)
-            if p is None:
-                log_debug("FAF package not found for {0}".format(str(ver)))
-            else:
-                log_debug("FAF package found for {0}".format(str(ver)))
-                if p.has_lob("package"):
-                    log_debug("LOB location {0}".format(p.get_lob_path("package")))
-                    return Path(p.get_lob_path("package"))
-                log_debug("LOB not found {0}".format(p.get_lob_path("package")))
-
-    # search for the debuginfo RPM
-    ver = None
-    repodir = Path(CONFIG["RepoDir"])
-    for release in repodir.iterdir():
-        for ver in vers:
-            testfile = release / "Packages" / ver.package_name(debug=True)
-            log_debug("Trying debuginfo file: %s" % testfile)
-            if testfile.is_file():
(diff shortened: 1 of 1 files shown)


DaveWysochanskiRH:
I guess this probably fixes this traceback:
~~~
  File "/usr/share/retrace-server/manager.wsgi", line 10, in <module>
    from retrace.retrace import (STATUS, STATUS_DOWNLOADING, STATUS_FAIL,
   File "/usr/lib/python3.6/site-packages/retrace/__init__.py", line 6, in <module>
     from . import retrace
   File "/usr/lib/python3.6/site-packages/retrace/retrace.py", line 394, in <module>
     def find_kernel_debuginfo(kernelver: KernelVer) -> Optional[Path]:
 NameError: name 'KernelVer' is not defined
~~~

michalfabik:
> I guess this probably fixes this traceback:
> 
> ```
> Traceback (most recent call last):
>   File "/usr/share/retrace-server/manager.wsgi", line 10, in <module>
>     from retrace.retrace import (STATUS, STATUS_DOWNLOADING, STATUS_FAIL,
>    File "/usr/lib/python3.6/site-packages/retrace/__init__.py", line 6, in <module>
>      from . import retrace
>    File "/usr/lib/python3.6/site-packages/retrace/retrace.py", line 394, in <module>
>      def find_kernel_debuginfo(kernelver: KernelVer) -> Optional[Path]:
>  NameError: name 'KernelVer' is not defined
> ```
  
  
Yeah, that was the one.

mzidek-gh:
Can look at the results from mypy test? It looks like this patch introduced an issue there.

ernestask:
> Can look at the results from mypy test? It looks like this patch introduced an issue there.

You might want to be more specific, because the results contain plenty of things to fix even before these changes.

mzidek-gh:
For convenience here are the issues detected by mypy:

```
Generated files under mypy_stubs/pyfaf/
build/src/retrace/config.py: note: In class "__config":
build/src/retrace/config.py:27:9: error: Need type annotation for 'ARCH_HOSTS'
(hint: "ARCH_HOSTS: Dict[<type>, <type>] = ...")  [var-annotated]
            ARCH_HOSTS = {}
            ^
mypy_stubs/pyfaf/storage/__init__.pyi:13: note: In module imported here,
build/src/retrace/retrace.py:351: note: ... from here:
mypy_stubs/pyfaf/storage/report.pyi:10:1: error: Module 'pyfaf.utils.parse' has
no attribute 'signal2name'  [attr-defined]
    from pyfaf.utils.parse import signal2name as signal2name
    ^
build/src/retrace/retrace.py: note: In function "run_gdb":
build/src/retrace/retrace.py:288:17: error: Incompatible types in assignment
(expression has type "CompletedProcess[str]", variable has type
"CompletedProcess[bytes]")  [assignment]
            child = run(["/usr/bin/mock", "chroot", "--configdir", savedir...
                    ^
build/src/retrace/retrace.py:310:21: error: Argument 1 to "run" has
incompatible type "List[object]"; expected
"Union[Union[bytes, str], Sequence[Union[bytes, str, _PathLike[Any]]]]" 
[arg-type]
            child = run(podman_build_call, stdout=DEVNULL, stderr=DEVNULL)
                        ^
build/src/retrace/retrace.py:314:17: error: Incompatible types in assignment
(expression has type "CompletedProcess[str]", variable has type
```

I see that we had mypy failures in previous PRs as well, but I those were different.

mzidek-gh:
Ah, wait they are actually not new. Please ignore the request for change.

mzidek-gh:
Ok, so it does not introduce but actually fixes a bunch of the issues detected by mypy (my bad, sould read the logs better).

In terms of coding style this just moves the code around so the code is not changed in that regard.

It compiles fine. As this is not a bug fix per se, there is no need for automated tests or anything like that to be added.

mzidek-gh:
ACK.

PR #350, opened 2020-07-23
Add back symlink to retrace-server-task.txt

ernestask:
3f46464be80fecfb2c1cda262f2fdd3ad23f3ac2 removed the symlink to
man/retrace-server-task.txt, which was added to retain compatibility for
Autotools builds.


DaveWysochanskiRH:
Thanks - not sure how / why I did that!

Issue #347, opened 2020-06-29
After 020ffa6442a8 queueing vmem + vmss file fails with "argument of type 'PosixPath' is not iterable"


DaveWysochanskiRH:
After the latest commit 020ffa6442a8, queuing any vmcore seems to be ok, but if I try to queue a vmem file with 'vmss' it fails with 
~~~
[2020-06-29 12:32:31] [I] Downloading remote resources
[2020-06-29 12:32:31] [I] Retrieving local file '/cores/bugzilla/regression-test/rhel7/vmware.vmem'
[2020-06-29 12:32:31] [D] File type: data
[2020-06-29 12:32:31] [D] unknown file type, unpacking finished
[2020-06-29 12:32:31] [D] Trying hardlink
[2020-06-29 12:32:31] [D] Succeeded
[2020-06-29 12:32:31] [I] Calculating md5sum
[2020-06-29 12:34:01] [I] Post-processing downloaded file
[2020-06-29 12:34:01] [D] File type: data
[2020-06-29 12:34:01] [D] unknown file type, unpacking finished
[2020-06-29 12:34:01] [D] File type: data
[2020-06-29 12:34:01] [D] unknown file type, unpacking finished
[2020-06-29 12:34:01] [I] Downloading remote resources
[2020-06-29 12:34:01] [I] Retrieving local file '/cores/bugzilla/regression-test/rhel7/vmware.vmss'
[2020-06-29 12:34:01] [D] File type: diy-thermocam raw data (lepton 2.x), scale 28263-29440, spot sensor temperature 0.000000, unit celsius, color scheme 0, calibration: offset 0.000000, slope 18179583481275050988767358746624.000000
[2020-06-29 12:34:01] [D] unknown file type, unpacking finished
[2020-06-29 12:34:01] [D] Trying hardlink
[2020-06-29 12:34:01] [D] Succeeded
[2020-06-29 12:34:01] [I] Calculating md5sum
[2020-06-29 12:34:01] [I] Post-processing downloaded file
[2020-06-29 12:34:01] [D] File type: diy-thermocam raw data (lepton 2.x), scale 28263-29440, spot sensor temperature 0.000000, unit celsius, color scheme 0, calibration: offset 0.000000, slope 18179583481275050988767358746624.000000
[2020-06-29 12:34:01] [D] unknown file type, unpacking finished
[...]

PR #352, opened 2020-07-26
Fix vmware snapshots

DaveWysochanskiRH:
This fixes the first two problems with vmware snapshots recently introduced.  The next problem has something to do with one of the two files being renamed / removed and results in the following error:
[2020-07-25 22:37:01] [E] [Errno 2] No such file or directory: '/cores/retrace/tasks/651051598/crash/vmware.vmss' -> '/cores/retrace/tasks/651051598/crash/vmcore/.vmss'

### DIFF
--- a/src/retrace/retrace.py
+++ b/src/retrace/retrace.py
@@ -452,13 +452,13 @@ def get_archive_type(path: Union[str, Path]) -> int:
     log_debug("unknown file type, unpacking finished")
     return ARCHIVE_UNKNOWN
 
-def add_snapshot_suffix(filename: str, snapshot: Path) -> str:
+def add_snapshot_suffix(filename: Path, snapshot: Path) -> str:
     """
     Adds a snapshot suffix to the filename.
     """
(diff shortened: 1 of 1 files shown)


DaveWysochanskiRH:
Now unpack_vmcore is failing which was not too easy to track down due to how except statements are now done.  To be honest, I'm fairly lost as to what this code intent was trying to do, and how it all broke, and I've spent too much time on this.  It looks like a variety of code paths have been broken since April 1st when that original Path commit went in.  Due to other changes, I could not merge to upstream until about a month ago when I found out this was broken.  Since then I've been just pulling in commits as I need them.  Maybe I should just cool it for now and go back to what I was doing in my own fork of upstream until this area can be looked at a bit more and until https://github.com/abrt/retrace-server/issues/347 is resolved fully.  Looking at unpack_vmcore and some of the other functions, possibly this area needs more work and rethought but not sure.

mgrabovsky:
Closing this as I take over in #356.

Issue #354, opened 2020-08-01
crash commands that timeout due to hitting ProcessCommunicateTimeout throw cryptic error: "[E] not all arguments converted during string formatting" error


DaveWysochanskiRH:
Occasionally we have runaway crash commands and apparently there is now a bug in upstream crash again that causes "crash --minimal" to spin at 100% CPU indefinitely on seemingly any vmcore.  This used to throw an error message to the log as defined in run_crash_cmdline().  Somewhere along the way this got broken maybe the patch that convered Popen to run or some other patch?  This isn't something that happens too often, but we put this in so we would not need a ticket to kill a runaway crash - the error message is important in that regard.  It's hard to notice these types of regressions so I'm not sure when it happened.
~~~
1314     def run_crash_cmdline(self, crash_start, crash_cmdline):
1315         cmd_output = None
1316         returncode = 0
1317         crashdir = self.get_crashdir()
1318         try:
1319             t = 3600
1320             if CONFIG["ProcessCommunicateTimeout"]:
1321                 t = CONFIG["ProcessCommunicateTimeout"]
1322             child = run(crash_start, stdout=PIPE, stderr=STDOUT,
1323                         cwd=crashdir, timeout=t,
1324                         input=crash_cmdline.encode())
1325             cmd_output = child.stdout
1326         except OSError as err:
1327             log_warn("crash command: '%s' triggered OSError " %
1328                      crash_cmdline.replace('\r', '; ').replace('\n', '; '))
1329             log_warn("  %s" % err)
1330         except TimeoutExpired:
1331             raise Exception("WARNING: crash command: '%s' exceeded " + str(t) +
1332                             " second timeout - damaged vmcore?" %
1333                             crash_cmdline.replace('\r', '; ').replace('\n', '; '))
~~~

PR #355, opened 2020-08-01
retrace: fix timeout exception error message in run_crash_cmdline

DaveWysochanskiRH:
Somewhere along the line the Exception string failed to format
properly and the following cryptic error message would be thrown
on a crash command that exceeded ProcessCommunicateTimeout:
  [E] not all arguments converted during string formatting

After this patch we get back the more informative message:
  [E] WARNING: crash command: 'log; quit; ' exceeded 60  second timeout - damaged vmcore?


DaveWysochanskiRH:
Fixes https://github.com/abrt/retrace-server/issues/354

PR #356, opened 2020-08-04
More type annotations and error fixing

mgrabovsky:
Following up to #352

Resolves #347

PR #360, opened 2020-08-10
Fix some pylint issues

michalfabik:
This fixes whitespace issues I ran into while reviewing #356 and some other easy ones.  
  
See commits.

### DIFF
--- a/src/retrace/hooks/hooks.py
+++ b/src/retrace/hooks/hooks.py
@@ -12,23 +12,21 @@ from subprocess import PIPE, CalledProcessError, run, TimeoutExpired
 from retrace.retrace import log_info, log_error, log_debug, RetraceTask
 from .config import HOOK_PATH, HOOK_TIMEOUT, hooks_config
 
-"""
-    Hooks description:
-    pre_start -- When self.start() is called
-    start -- When task type is determined and the main task starts
-    pre_prepare_debuginfo -- Before the preparation of debuginfo packages
-    post_prepare_debuginfo -- After the preparation of debuginfo packages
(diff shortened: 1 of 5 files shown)


retrace-server.spec
```diff
@@ -1,6 +1,6 @@
 Summary: Application for remote coredump analysis
 Name: retrace-server
-Version: 1.21.0
+Version: 1.21.0.159+gc34c
```
xsuchy, line 3: Why this?

xsuchy:
I am afraid that the last commit ba31850c0ab82f07476e5155fcadbc5f827bf858 contains something totally different that the description states. :(

retrace-server.spec
michalfabik: Sorry, accident.

src/retrace/hooks/hooks.py
```diff
@@ -99,7 +99,7 @@ def _process_script(self, hook: str, hook_path: Path, exc_path: str) -> None:
 
         try:
             child = run(cmd, shell=True, timeout=hook_timeout, cwd=hook_path,
-                        stdout=PIPE, stderr=PIPE, encoding='utf-8')
+                        stdout=PIPE, stderr=PIPE, encoding='utf-8', check=False)
             child.check_returncode()
```
mgrabovsky, line 103: We could just shorten this to
```suggestion
                        stdout=PIPE, stderr=PIPE, encoding='utf-8', check=True)
```
since we're calling `check_returncode()` afterwards anyway.

michalfabik: Thanks, fixed.

Issue #364, opened 2020-08-24
retrace manager task page "Internal Error" and Traceback in /var/log/http/ssl_error_log if retrace_log contains non-ascii character


DaveWysochanskiRH:
We've seen a couple instances where attempting to display a retrace_log file causes a Traceback to manager page.  This happens because there's ways that non-ascii characters can get into the log file.
1. Use 'Custom core location' and leave out the first '/'.  Somehow 'wget' fails and then leaves a bit of garbage in the retrace_log file
2. Submit a tarball / file containing unicode characters from non-english country.

Example 2:
~~~
$ retrace-server-task create /tmp/127.0.0.1-2020-08-09-011828.tar.gz 
Status code 500 was returned with message: '<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>500 Internal Server Error</title>
</head><body>
<h1>Internal Server Error</h1>
<p>The server encountered an internal error or
misconfiguration and was unable to complete
your request.</p>
<p>Please contact the server administrator at 
 root@localhost to inform them of the time this error occurred,
 and the actions you performed just before this error.</p>
<p>More information about this error may be available
in the server error log.</p>
</body></html>
'
~~~

Log shows
~~~
tail /var/log/httpd/ssl_error_log | cut -d\  -f12-
Traceback (most recent call last):
  File "/usr/share/retrace-server/manager.wsgi", line 456, in application
    backtracewindow = "<h2>Log:</h2><textarea class=\\"backtrace\\">%s</textarea>" % task.get_log()
  File "/usr/lib/python3.6/site-packages/retrace/retrace.py", line 1171, in get_log
    return self.get(RetraceTask.LOG_FILE, maxlen=1 << 22)
  File "/usr/lib/python3.6/site-packages/retrace/retrace.py", line 1082, in get
    result = f.read(maxlen)
  File "/usr/lib64/python3.6/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xef in position 159: ordinal not in range(128)
~~~

PR #365, opened 2020-09-02
retrace: Open log files with utf-8 encoding

mzidek-gh:
We do not specify encoding when reading the log file. This causes
https://github.com/abrt/retrace-server/issues/364

I wonder if we should use utf-8 hardcoded or first try the environment settings and fall back to utf-8 if it does not work.

What do you guys think?

### DIFF
--- a/src/retrace/retrace.py
+++ b/src/retrace/retrace.py
@@ -1078,7 +1078,7 @@ class RetraceTask:
             return None
 
         filename = self._get_file_path(key)
-        with open(filename, "r") as f:
+        with open(filename, "r", encoding='utf-8', errors='replace') as f:
             result = f.read(maxlen)
 
         return result


DaveWysochanskiRH:
Confirmed this fixes the issue.  
I would say use 'utf-8' hardcoded since I would view try/except as unnecessarily complex.

mzidek-gh:
Thanks for the testing @DaveWysochanskiRH .

I am also leaning towards hard-coding the utf-8 value as the try-except version looks much uglier.

DaveWysochanskiRH:
> Thanks for the testing @DaveWysochanskiRH .
> 
Sure thing!

> I am also leaning towards hard-coding the utf-8 value as the try-except version looks much uglier.

It's good to try alternative approaches.

src/retrace/retrace.py
```diff
@@ -1078,7 +1078,7 @@ def get(self, key: Union[str, Path], maxlen: int = 268435456) -> Optional[str]:
             return None
 
         filename = self._get_file_path(key)
-        with open(filename, "r") as f:
+        with open(filename, "r", encoding='utf-8') as f:
```
ernestask, line 1081: Should we also pass `errors` with `replace` or something other as the value? Just in case of corruption.

mzidek-gh: idk, would it not be better to fail in case of corruption?

ernestask: Are we failing in a controlled manner or just giving the user an unhandled exception?

ernestask: It would be better to have things just work. I don’t know what the consequences would be of showing replacement characters in cases where something magically fails.

mzidek-gh: Ok, I think you are right. I added the errors='replace' to the open().

DaveWysochanskiRH: I like the addition of 'errors' handler.  In this case, I think errors='ignore' is better than errors='replace' though, unless we actually want garbage / non-ascii characters printed?

I tested both errors='ignore' and errors='replace' with an inserted 0xff character in one retrace_server log file and then using a web browser to dump the retrace_log (via manager interface).  As expected, 'ignore' does not print anything but 'replace' prints a garbage character at the end.  If you don't have an errors= then this gives a Traceback internally and "Internal server error" on the manager interface / browser.

DaveWysochanskiRH: FWIW, reference to the handlers - https://docs.python.org/3/library/codecs.html#error-handlers

DaveWysochanskiRH: If you wanted to get fancy there may be some merit in some sort of exception handling and printing to the httpd log the location of the bad bytes as would occur if there were no 'errors=' on the open, or a 'try .. except' clause:
~~~
error decoding file /cores/retrace/tasks/497308754/retrace_log: 'utf-8' codec can't decode byte 0xff in position 429: invalid start byte
~~~

The above may be overkill, but this is the 'get()' method so it handles a lot of files, though most should not have problems with non-ascii chars. In my experience decoding errors have only occurred when some other programs output is dumped into a file (in this case, wget for whatever reason put a non-ascii character in there, and in one other case 'crash' output was non-ascii).

mzidek-gh: I really do not like loosing information :) Replacing a non ascii character with <?> at least tells us that there are non-ascii characters. If we just ignore them we will loose this information. Now, I do not think it is very valuable information in most cases, but it may be in some. Unless the replace method causes issues (IIUC it does not) I would go with that.

ernestask: > In my experience decoding errors have only occurred when some other programs output is dumped into a file (in this case, wget for whatever reason put a non-ascii character in there, and in one other case 'crash' output was non-ascii).

Bingo. Note that I intentionally did not dig to see if there are scenarios where we could get garbage without cosmic rays being involved, so it might very well be just a precaution, but it seems like a cheap way to prevent something like that without erroring out, so why not.

ernestask (approved):

Issue #367, opened 2020-09-09
retrace-server-task Tracebacks if no kerberos ticket and retrace-server requires kerberos


DaveWysochanskiRH:
The retrace-server-task commands will traceback if the server requires kerberos but there's no kerberos ticket.  Once a user kinits the command goes through fine.

Example (retrace-server-task restart):
~~~
$ retrace-server-task restart 582204476
Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/urllib3/connection.py", line 159, in _new_conn
    (self._dns_host, self.port), self.timeout, **extra_kw)
  File "/usr/lib/python3.6/site-packages/urllib3/util/connection.py", line 57, in create_connection
    for res in socket.getaddrinfo(host, port, family, socket.SOCK_STREAM):
  File "/usr/lib64/python3.6/socket.py", line 745, in getaddrinfo
    for res in _socket.getaddrinfo(host, port, family, type, proto, flags):
socket.gaierror: [Errno -2] Name or service not known

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/bin/retrace-server-task", line 721, in <module>
    restart_existing_task(ARGS)
  File "/usr/bin/retrace-server-task", line 170, in restart_existing_task
    auth=args['kerberos'])
  File "/usr/lib/python3.6/site-packages/requests/api.py", line 116, in post
    return request('post', url, data=data, json=json, **kwargs)
  File "/usr/lib/python3.6/site-packages/requests/api.py", line 60, in request
    return session.request(method=method, url=url, **kwargs)
  File "/usr/lib/python3.6/site-packages/requests/sessions.py", line 533, in request
    resp = self.send(prep, **send_kwargs)
  File "/usr/lib/python3.6/site-packages/requests/sessions.py", line 646, in send
    r = adapter.send(request, **kwargs)
  File "/usr/lib/python3.6/site-packages/requests/adapters.py", line 467, in send
    low_conn.endheaders()
  File "/usr/lib64/python3.6/http/client.py", line 1249, in endheaders
    self._send_output(message_body, encode_chunked=encode_chunked)
  File "/usr/lib64/python3.6/http/client.py", line 1036, in _send_output
    self.send(msg)
  File "/usr/lib64/python3.6/http/client.py", line 974, in send
    self.connect()
  File "/usr/lib/python3.6/site-packages/urllib3/connection.py", line 301, in connect
    conn = self._new_conn()
[...]

PR #369, opened 2020-09-23
Fix crash if Kerberos authentication fails

mgrabovsky:
If `retracer-server-task` cannot connect to the (default) `localhost` server, it falls back to retrace.fedoraproject.org, but does not check if it's able to connect there.

The problem is that it falls back to retrace.fp.org if the connection fails for any reason. Specifically, it falls back even if authentication fails on localhost.

Moreover, if retrace.fp.org is inacessible as well, we get an exception later down the road when trying to send an HTTP request.

In this change, we check if the server responded HTTP 401 Unauthorized, which should be the case if a valaid Kerberos ticket was expected by the server but not sent by the client.

Plus fix some code style issues reported by Pylint.

### DIFF
--- a/src/retrace-server-task
+++ b/src/retrace-server-task
@@ -21,6 +21,7 @@ from requests_gssapi import HTTPSPNEGOAuth, DISABLED
 from retrace.retrace import add_snapshot_suffix
 
 
+FALLBACK_SERVER = 'https://retrace.fedoraproject.org'
 TASK_RETRACE, TASK_DEBUG, TASK_VMCORE, TASK_RETRACE_INTERACTIVE, \
     TASK_VMCORE_INTERACTIVE = range(5)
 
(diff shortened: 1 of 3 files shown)


mgrabovsky:
@DaveWysochanskiRH Can you please verify this fixes the issue in #367 for you?

DaveWysochanskiRH (approved):

DaveWysochanskiRH (approved):
LGTM

src/retrace-server-task
```diff
@@ -21,8 +21,9 @@ from requests_gssapi import HTTPSPNEGOAuth, DISABLED
 from retrace.retrace import add_snapshot_suffix
 
 
+FALLBACK_SERVER = 'https://retrace.fedoraproject.org'
```
mkutlak, line 24: Can you make this configurable?

xsuchy (approved):

Issue #374, opened 2020-10-01
RFE: Add 'restart task' button on the manager task page similar to 'delete task'


DaveWysochanskiRH:
This probably should depend on https://github.com/abrt/retrace-server/issues/373
Now that we have the 'restart' option of retrace-server-task, it probably would be fairly straightforward to implement a 'restart' button the 'manager' page, but may require some refactoring.  This was something that was asked for quite a while back, since many use the 'manager' page.

The design I'm thinking is as follows:
1. Add a 'restart task' link that only shows up on a specific tasks' page if the task is failed.  This would be similar to the 'Delete task' link.
2. If the 'restart task' link is followed, open a new page which would be very similar, if not identical to the page when an 'ftp' task is selected.  On that page there are options to fill in kernelver, caseno, bugzillano, md5sum, etc

Issue #373, opened 2020-10-01
RFE: Improve retrace-server-task restart to include 'caseno', 'bugzillano', 'email' and all options on the ftp 'create task' manager page


DaveWysochanskiRH:
The commit 6bf02a85 that implemented "restart" option of retrace-server-task (and man page update commit 3f46464be8) neglected to include a few common options given with the "create" command, which should not be too hard to implement.  The options needed are ones that would enable a 'restart task' link (see https://github.com/abrt/retrace-server/issues/374) similar to the page when a new ftp task is created.  The options as seen on the 'create remote ftp task' page are as follows:
* Kernel version (empty to autodetect):
* Case no.
* Bugzilla no.
* E-mail notification
* Be more verbose in case of error
* Calculate md5 checksum for all downloaded resources

DaveWysochanskiRH:
Probably also need --vmem option - does not look like restarting a task with vmem works.

PR #375, opened 2020-10-01
retrace-server-task: Add options to restart command similar to create

DaveWysochanskiRH:
Commit 6bf02a8 implemented a 'restart' option on the retrace-server-task
command, but did not implement all options which would be appropriate
for restart.  Add options options email, caseno, bugzillano, and md5sum.

Since these options are the same as when we 'start' a new task, refactor
the manager code on the backend to avoid parsing duplication.

### DIFF
--- a/man/retrace-server-task.txt
+++ b/man/retrace-server-task.txt
@@ -7,23 +7,25 @@ retrace-server-task - Create a task, get and set fields on an existing task
 
 SYNOPSIS
 --------
-'retrace-server-task' create [-h] [-u | -k] [-t | -m | -f] [-i] [-d]
-                             [-s SERVER] [-n] [-v] [--vmem VMEM] [-r KERNELVER]
+'retrace-server-task' create [-h] [-n] [-s SERVER] [-v]
+                             [-u | -k] [-t | -m | -f] [-i] [-d]
+                             [-bz BUGZILLANO] [-c CASENO] [-e EMAIL]
                              [-o OS_RELEASE] [-p PACKAGE] [-x EXECUTABLE]
-                             [-e EMAIL] [-c CASENO] [-bz BUGZILLANO]
+                             [-r KERNELVER]
+                             [--vmem VMEM]
                              COREFILE
 
-'retrace-server-task' restart [-h] [-s SERVER] [-n] [-v]
+'retrace-server-task' restart [-h] [-n] [-s SERVER] [-v]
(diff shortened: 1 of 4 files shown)


DaveWysochanskiRH:
This is initial implementation which seems to work so far based on my testing.  So far I have tested these options on both restart and create but only on vmcores.  Probably more refactoring could be done inside retrace-server-task due to the common options though I'm not sure how far to go.  The 'create' option for example actually makes separate server calls to set the email, caseno, bugzillano rather than passing into the task creation.  I'm not sure I should that should be added here but maybe should be done for code sharing reasons.

DaveWysochanskiRH:
Is there a need to have restart for coredumps (userspace cores)?  Unfortunately I still do have these turned on in our retrace-server so I cannot test these or know if this is something that is needed even.

DaveWysochanskiRH:
Commit 1f6eacece6cd takes a stab at https://github.com/abrt/retrace-server/issues/374 and it does actually restart the task via the manager page, however the parameters such as 'bugzillano', 'email', etc are not passed properly so those don't work yet.  I'm a little unsure about this since with some methods the parameters are passed in the URL, and others they are passed in data of a POST operation (this is what works today).  Maybe this is the difference between 'GET' and 'POST' http ops, and maybe the approach in commit 1f6eacece6cd is wrong.

DaveWysochanskiRH:
I just pushed my latest patches for this which should address both RFEs I opened about restarting tasks:
* https://github.com/abrt/retrace-server/issues/373
* https://github.com/abrt/retrace-server/issues/374

I'm testing now and they seem to work ok.  Some of the manager forms code is a bit awkward though as I tried to avoid duplication, though I'm not sure I ended up at a good place.  Probably more refactoring could be done for the forms, possibly another round of patches is warranted.

DaveWysochanskiRH:
Ok I found another problem, so I'm definitely going to remove get_startcontent() which was an attempt at avoiding duplication of the manager page form code.  Probably get_startcontent() tried to do too much.  Possible I can create smaller helpers, but I'm probably going to just start with a form for 'restart' and avoid touching the 'start' form.

DaveWysochanskiRH:
Ok this only addresses the additional options to the retrace-server-task command:
https://github.com/abrt/retrace-server/issues/373

The other RFE I think needs more work, but depends on this one.

michalfabik (approved):

PR #376, opened 2020-10-05
retrace-server-reposync: Fix path concatenation error

michalfabik:
`retrace-server-reposync` was failing on a misused `/` (path concatenation operator) due to dir keys in `CONFIG` being ordinary strings, not `Path`s. This commit fixes this and removes one redundant use of `/`, for more consistency.


mgrabovsky (approved):
Good catch.

PR #383, opened 2020-10-15
create: Fix type error

mgrabovsky:
This fixes the error "Crash directory does not contain required file 'coredump'".

The bug was introduced in 94a71c2 during the transition to pathlib.


michalfabik (approved):

mzidek-gh (approved):

mzidek-gh:
ACK.

PR #387, opened 2020-10-21
hooks: Change path to python interpreter from /bin to /usr/bin

DaveWysochanskiRH:
Install was giving this error which indicated /bin/python3 was
not provided:
Error:
 Problem: conflicting requests
  - nothing provides /bin/python3 needed by retrace-server-1.21.0-1.git.1011.3e498e3.el8.noarch
(try to add '--skip-broken' to skip uninstallable packages or '--nobest' to use not only best candidate packages)

### DIFF
--- a/src/backtrace.wsgi
+++ b/src/backtrace.wsgi
@@ -1,4 +1,3 @@
-#!/usr/bin/python3
 from webob import Request
 
 from retrace.retrace import RetraceTask
--- a/src/checkpackage.wsgi
+++ b/src/checkpackage.wsgi
@@ -1,4 +1,3 @@
-#!/usr/bin/python3
 from webob import Request
 
 from retrace.retrace import is_package_known
--- a/src/create.wsgi
+++ b/src/create.wsgi
@@ -1,4 +1,3 @@
-#!/usr/bin/python3
 import os
 import sys
 from typing import List
--- a/src/delete.wsgi
+++ b/src/delete.wsgi
@@ -1,4 +1,3 @@
-#!/usr/bin/python3
 from webob import Request
 
 from retrace.retrace import RetraceTask
--- a/src/exploitable.wsgi
+++ b/src/exploitable.wsgi
@@ -1,4 +1,3 @@
-#!/usr/bin/python3
 from webob import Request
 
 from retrace.retrace import RetraceTask
--- a/src/ftp.wsgi
+++ b/src/ftp.wsgi
@@ -1,5 +1,3 @@
-#!/usr/bin/python3
-
 import fnmatch
 import re
 import urllib
--- a/src/index.wsgi
+++ b/src/index.wsgi
@@ -1,4 +1,3 @@
-#!/usr/bin/python3
 from webob import Request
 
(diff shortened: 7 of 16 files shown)


DaveWysochanskiRH:
After pulling the latest merge of https://github.com/abrt/retrace-server/pull/385 I converted to tito building.  After that, the install step was failing due to this one path of "/bin/python3", which is the only place in the source that has this path.  This small patch fixed the install error.

mgrabovsky:
Actually, the file should not have a shebang at all, since it's not an executable. So it shouldn't be +x either (which it is right now). There's also `src/retrace/hooks/hooks.py` with the same issue.

Could you please update the PR with these changes?

DaveWysochanskiRH:
> Actually, the file should not have a shebang at all, since it's not an executable. So it shouldn't be +x either (which it is right now). There's also `src/retrace/hooks/hooks.py` with the same issue.
> 
> Could you please update the PR with these changes?

Ok will do!

DaveWysochanskiRH:
@mgrabovsky - What about argparser?
./src/retrace/argparser.py:#!/usr/bin/python3

mgrabovsky:
Yep, same thing. Good catch. I was going after the executable flag initially.

DaveWysochanskiRH:
> Yep, same thing. Good catch. I was going after the executable flag initially.
> 
> Edit: Now that I'm looking into it, there are also several `.wsgi` scripts that match the pattern.

Ok do we need a python path at all?  Are we going to run into problems with systems that might have python2 as the interpreter?

DaveWysochanskiRH:
I think this is right - testing now...
~~~
commit 8bc7c3c36d6c279d6526e5c6513e8cf7c9d63c4b
Author: Dave Wysochanski <dwysocha@redhat.com>
Date:   Wed Oct 21 11:38:25 2020 -0400

    wsgi: Remove python3 interpreter paths in all files
    
    An WSGI application is a python application.  Since we already
    have build dependencies on python3, there's no need to specify
    the path to the interpreter in these files.
    
diff --git a/src/backtrace.wsgi b/src/backtrace.wsgi
index ed19d6903935..48faf5dc1993 100644
--- a/src/backtrace.wsgi
+++ b/src/backtrace.wsgi
@@ -1,4 +1,3 @@
-#!/usr/bin/python3
 from webob import Request
 
 from retrace.retrace import RetraceTask
diff --git a/src/checkpackage.wsgi b/src/checkpackage.wsgi
index ecc76444dba9..bcba825214af 100644
--- a/src/checkpackage.wsgi
+++ b/src/checkpackage.wsgi
@@ -1,4 +1,3 @@
-#!/usr/bin/python3
 from webob import Request
 
~~~

mgrabovsky:
I think it's okay to drop it in all of them, since they're not executables. [Fedora Packaging Guidelines](https://docs.fedoraproject.org/en-US/packaging-guidelines/#_shebang_lines) state that non-executables should not have shebangs.

mgrabovsky (approved):
Looks great. Thanks!

PR #388, opened 2020-10-21
Support testing via Meson

mgrabovsky:
The Meson script works, but I'm not sure it follows Meson's best practices. The test is currently broken.

Follow-up to #385.


mgrabovsky:
Ready for review now.

The test now goes through locally (at least on my machine). You can run it as follows:
```bash
meson . _build
meson test -C _build --print-errorlogs
```

(Note that several packages are required for the test to run, which Meson will helpfully inform you of.)

In the end, I chose not to run it during RPM build (in the spec file). One reason is that it requires a network connection (in order to pull packages) which might be a problem in Copr or Koji. It also makes the test sort of unreliable/unpredictable both functionally and in timing.

Another reason is that, when running the build in Mock locally, I get the following error from Podman when retrace server is attempting to build the image required for retracing:
> Error: cannot setup namespace using newuidmap

I suppose we could revamp the test completely, create multiple unit tests etc. However, with retrace server potentially retiring soon-ish, I find that unnecessary.

xsuchy:
@michalfabik can you review this please<

michalfabik:
Having an older version of `retrace-server` installed seems to be causing the tests to load old libraries, potentially making the tests fail.

mgrabovsky:
@michalfabik I have fixed the problem, although I think it's advisable to run the tests in Mock or a container anyway. (Same way it's done in Koji, for instance.)

michalfabik:
LGTM

michalfabik (approved):

PR #389, opened 2020-10-21
add releasers.conf

xsuchy:
This way you can just 'tito release all' or 'tito release epel'.

@mgrabovsky Can you build it for epel too?


mgrabovsky (approved):
Looks good to me.

PR #392, opened 2020-10-23
Fix path to coredump in dockerfile

### LINKED ISSUE #391: Podman image build failure
When using the Podman backend, the image build fails with the following error on retrace.fp.org:
```
STEP 1: FROM fedora:32
STEP 2: RUN useradd -l retrace &&     mkdir -p /var/spool/abrt/crash
useradd: /etc/passwd.7: lock file already used
useradd: cannot lock /etc/passwd; try again later.
Error: error building at STEP "RUN useradd -l retrace &&     mkdir -p /var/spool/abrt/crash": error while running runtime: exit status 1
```

I tried digging deeper what the problem might be, but with little success. All I found was containers/buildah#2239, which mentions some issues with RHEL 8.2 and fuse-overlayfs, but I don't know where to go from there.

mgrabovsky:
Resolves #391

### DIFF
--- a/src/retrace/retrace_worker.py
+++ b/src/retrace/retrace_worker.py
@@ -400,8 +400,9 @@ class RetraceWorker:
         self.hook.run("start")
 
         task = self.task
+        savedir = task.get_savedir()
         crashdir = task.get_crashdir()
-        corepath = crashdir / "coredump"
+        corepath = crashdir / RetraceTask.COREDUMP_FILE
 
         try:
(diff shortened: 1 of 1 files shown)


xsuchy (approved):

mzidek-gh:
Just a question. The changes in the mock are not relevant for the fix. Should we just remove the code?

mzidek-gh:
Nevermind that can be done later. ACK from me as well.

PR #395, opened 2020-10-30
Repair coredump2packages

mgrabovsky:
* Call the `main()` function in `coredump2packages`. The refactor in 81e750f failed to ever call the function.
* Only look for binary packages with same architecture as debuginfo packages.

This should fix many of the recent retracing failures.


mzidek-gh (approved):

mzidek-gh:
ACK.

PR #399, opened 2020-11-20
Fix reposync; update Fedora version list

mgrabovsky:
Second round of fixes for recent retrace server problems:
* Update Fedora plugin version list: Add F33 and remove versions older than F31.
* Fix type errors in `retrace-server-reposync-faf`.

### DIFF
--- a/src/plugins/fedora.py
+++ b/src/plugins/fedora.py
@@ -10,15 +10,6 @@ gpg_keys = [
     "/usr/share/distribution-gpg-keys/fedora/RPM-GPG-KEY-fedora-{release}-primary",
 ]
 versionlist = [
-    "fc22",
-    "fc23",
-    "fc24",
-    "fc25",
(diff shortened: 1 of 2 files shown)


mgrabovsky:
I think this can now be reviewed. I'll leave the Modularity stuff for a separate PR.

mzidek-gh:
Just one question. What does it mean for the user if fedora version is removed from the list? If I use fedora 30 after this patch, will I get some error message that the version is no longer supported?

mgrabovsky:
The version list is only used on the stats page, so it has little real-world effect.

mzidek-gh:
> The version list is only used on the stats page, so it has little real-world effect.

Ok, then I have no more comments. I have not tested the patches on a server, but I think they look safe enough to merge. So feel free to merge.

michalfabik (approved):

Issue #403, opened 2020-11-24
Missing dependency on logrotate


mgrabovsky:
[rpmlint](https://github.com/rpm-software-management/rpmlint) reports that retrace-server is missing a formal dependency on logrotate which should apparently be present due to the `/etc/logrotate.d/retrace-server` file.

xsuchy:
We should not require it, but rather recommend it. But we should own `/etc/logrotate.d/`.

mgrabovsky:
Indeed, [it can be](https://github.com/rpm-software-management/rpmlint/blob/c6b42fee4fd0534c7a7fa983f1e77bf5cf1ea2d2/rpmlint/checks/FilesCheck.py#L553-L555) either of `Requires`, `Recommends` or `Suggests`.

mgrabovsky:
Closed by #405.

PR #405, opened 2020-12-02
Update specfile: logrotate

mzidek-gh:
Add Recommneds for logrotate package.

PR #406, opened 2020-12-02
NEEDINFO: spec: Set ownership and permission for logotate.d

mzidek-gh:
According to a comment in:
https://github.com/abrt/retrace-server/issues/403

We should change the ownsership to our logrotate config file.

This change will result in both retrace and root (instead of just root) to be able to modify the config file.

I am not sure why retrace user needs to be able to do this. Usually configuration is only modofied by a user with root priviledges. Can someone clarify?

### DIFF
--- a/retrace-server.spec
+++ b/retrace-server.spec
@@ -171,7 +171,7 @@ exit 0
 %config(noreplace) %{_sysconfdir}/%{name}/hooks/start.conf
 %config(noreplace) %{_sysconfdir}/%{name}/hooks/success.conf
 %config(noreplace) %{_sysconfdir}/%{name}/hooks/task.conf
-%config(noreplace) %{_sysconfdir}/logrotate.d/%{name}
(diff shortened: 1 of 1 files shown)


retrace-server.spec
```diff
@@ -171,7 +171,7 @@ exit 0
 %config(noreplace) %{_sysconfdir}/%{name}/hooks/start.conf
 %config(noreplace) %{_sysconfdir}/%{name}/hooks/success.conf
 %config(noreplace) %{_sysconfdir}/%{name}/hooks/task.conf
-%config(noreplace) %{_sysconfdir}/logrotate.d/%{name}
```
xsuchy, line 174: It should be either

%{_sysconfdir}/logrotate.d/

or 

Requires: logrotate

mzidek-gh:
Ok, I completely misunderstood the issue. I will update the patch. Thanks for clarification.

mzidek-gh:
I have merged this PR with https://github.com/abrt/retrace-server/pull/405

Closing.

Issue #400, opened 2020-11-20
Handle crashes in modules


mgrabovsky:
retrace server currently cannot handle packages names (NVR) from crashes in modules. The following error is returned by retrace.fp.org when trying to retrace a crash in Node.js:
```
--- Running analyze_CCpp ---
Ok to upload core dump? (It may contain sensitive data). If your answer is 'No', a stack trace will be generated locally. (It may download a huge amount of data). 'YES'
Querying server settings
A server-side error occurred on 'https://retrace.fedoraproject.org'
Unexpected HTTP response from server: 403
Package NVR contains illegal characters
Do you want to generate a stack trace locally? (It may download a huge amount of data but reporting can't continue without stack trace). 'NO'
('analyze_CCpp' exited with 1)
```

The package version is `1:14.14.0-2.module_f33+10426+c7be8abe.x86_64`, which seems to upset [our NVR parser](https://github.com/abrt/retrace-server/blob/e2e6b95/src/checkpackage.wsgi#L21-L23). Later down the road, some adjustments might be necessary in [`is_package_known()`](https://github.com/abrt/retrace-server/blob/e2e6b95/src/retrace/retrace.py#L354), as well.

Steps to reproduce:
```
$ sudo dnf module install nodejs:14/default
$ node & sleep 1 && kill -ABRT $! && fg
$ abrt report
```

PR #409, opened 2020-12-16
Improve regexes for parsing problem data

mgrabovsky:
* Replace ranges with PCRE escapes, e.g. `a-zA-Z` → `\w`.
* Fix package name (NEV) parser to correctly handle epoch numbers. The epoch was previously assumed to be a prefix of the whole name (in ENV form), e.g. `1:findutils-4.7.0-4.fc33`. In order to be consistent with abrt, DNF and RPM, we switch to NEV form, that is `findutils-1:4.7.0-4.fc33`.

Potentially related: #400

### DIFF
--- a/src/retrace/util.py
+++ b/src/retrace/util.py
@@ -19,15 +19,15 @@ GETTEXT_DOMAIN = "retrace-server"
 DF_OUTPUT_PARSER = re.compile(r"^([^ ^\t]*)[ \t]+([0-9]+)[ \t]+([0-9]+)[ \t]+([0-9]+)[ \t]+([0-9]+%)[ \t]+(.*)$")
 
 # architecture (i386, x86_64, armv7hl, mips4kec)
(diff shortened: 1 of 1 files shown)


xsuchy:
I think we can do that as we run under en locale. But beware that e.g. in Swedish locale the "w" is not part of \w

xsuchy (approved):

mgrabovsky:
> I think we can do that as we run under en locale. But beware that e.g. in Swedish locale the "w" is not part of \w

Ah, I didn't think about that. But now that I'm looking into the [documentation](https://docs.python.org/3/library/re.html#index-32)`\w` includes “most characters that can be part of a word in any language, as well as numbers and the underscore”. So we should be fine regarding locales (although using the `ASCII` flag might be safer), but we're now accepting more than the original parser.

The change might have been too hasty. I will go and fix that.

PR #413, opened 2021-02-17
retrace: Fix "file not found" when the archive contains directories

pierg75:
When iterating over some directories found in the crashdir,
retrace was trying to move the "fullpath" (i.e. the original
file) instead of "file_path" (the directory), and the task
would fail with the following message:

[2021-02-04 10:52:48] [E] Task failed: [Errno 2] No such file or
directory: '/cores/retrace/tasks/231067672/crash/testfile-vmcore'

Signed-off-by: Pierguido Lambri <plambri@redhat.com>
Reviewed-by: Dave Wysochanski <dwysocha@redhat.com>

### DIFF
--- a/src/retrace/retrace.py
+++ b/src/retrace/retrace.py
@@ -1514,7 +1514,7 @@ class RetraceTask:
             vmcore_path = self.get_vmcore_path()
             for file_path in crashdir.iterdir():
                 if file_path.is_dir():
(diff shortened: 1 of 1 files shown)


DaveWysochanskiRH (approved):
Pier did the fix but I helped him test and it looks good to me.

mgrabovsky (approved):
Looks right to me. Thanks for the fix.

PR #418, opened 2021-04-19
Podman backend refactor and fixes

mgrabovsky:
These changes entail a complete refactoring of the Podman-based retracing backend. The biggest change is that only one container image is now built for each release (e.g. Fedora 32, Fedora 33, etc.) with basic tools preinstalled (abrt-addon-ccpp and GDB) and new containers are run from this base image on demand.

Some more changes are made for improving readability and maintainability.

See commits for details.

This code is currently deployed and ready for experimenting at <https://retrace-stg.aws.fedoraproject.org/> (only Fedora 33 supported).

Should fix at least [rhbz#1922471](https://bugzilla.redhat.com/show_bug.cgi?id=1922471) and [rhbz#1932768](https://bugzilla.redhat.com/show_bug.cgi?id=1932768).

### DIFF
--- a/retrace-server.spec
+++ b/retrace-server.spec
@@ -73,8 +73,9 @@ Requires(post): /sbin/install-info
 %endif
 Requires(post): /usr/bin/crontab
 Requires(post): /usr/bin/systemctl
-Recommends: podman
+Recommends: httpd
 Recommends: logrotate
+Recommends: podman
 
 Obsoletes: abrt-retrace-server < 2.0.3
 Provides: abrt-retrace-server = 2.0.3
--- a/retrace-server.spec
+++ b/retrace-server.spec
@@ -1,3 +1,7 @@
+#retrace uid/gid reserved in setup, rhbz #706012
+%define retrace_gid_uid 174
+%define retrace_home %{_sharedstatedir}/retrace
+
 Summary: Application for remote coredump analysis
 Name: retrace-server
 Version: 1.23.0
@@ -68,6 +72,7 @@ Requires(preun): /sbin/install-info
 Requires(post): /sbin/install-info
 %endif
 Requires(post): /usr/bin/crontab
+Requires(post): /usr/bin/systemctl
 Recommends: podman
 Recommends: logrotate
 
@@ -131,10 +136,8 @@ rm -f %{buildroot}%{_infodir}/dir
 %{find_lang} %{name}
 
 %pre
-#retrace uid/gid reserved in setup, rhbz #706012
-%define retrace_gid_uid 174
 getent group retrace > /dev/null || groupadd -f -g %{retrace_gid_uid} --system retrace
-getent passwd retrace > /dev/null || useradd --system -g retrace -u %{retrace_gid_uid} -d %{_sharedstatedir}/retrace -s /sbin/nologin retrace
+getent passwd retrace > /dev/null || useradd --system -g retrace -u %{retrace_gid_uid} -d %{retrace_home} -s /sbin/nologin retrace
 exit 0
 
 %post
@@ -143,16 +146,28 @@ exit 0
 %endif
 /usr/sbin/usermod -a -G mock retrace 2> /dev/null || :
 
+# Migrate old home directory during an upgrade (i.e. $1 == 2) -- if retrace's home
+# is /usr/share/retrace change it to /var/lib/retrace.
+if [ "$1" = 2 -a "$(getent passwd retrace | cut -d: -f6)" = "%{_datadir}/%{name}" ]
+then
+    # Make sure to restart the Apache HTTP Server if it's enabled and
+    # running during the upgrade.
+    /usr/bin/systemctl is-active --quiet httpd.service ; httpd_active=$?
+    [ "$httpd_active" = 0 ] && /usr/bin/systemctl stop httpd.service
+    /usr/sbin/usermod --home %{retrace_home} retrace 2> /dev/null || :
+    [ "$httpd_active" = 0 ] && /usr/bin/systemctl start httpd.service
+fi
+
 if [ "$1" = 1 ]
 then
 #add disabled crontab entries to retrace's crontab
     %define retrace_crontab_entry0 "# 0 * * * * /usr/bin/retrace-server-cleanup >> /var/log/retrace-server/cleanup_error.log 2>&1"
-    %define retrace_crontab_entry1 "#0 0,12 * * * /usr/bin/retrace-server-reposync fedora 15 i386 >> /var/log/retrace-server/reposync_error.log 2>&1"
-    %define retrace_crontab_entry2 "#0 2,14 * * * /usr/bin/retrace-server-reposync fedora 15 x86_64 >> /var/log/retrace-server/reposync_error.log 2>&1"
-    %define retrace_crontab_entry3 "#0 4,16 * * * /usr/bin/retrace-server-reposync fedora 16 i386 >> /var/log/retrace-server/reposync_error.log 2>&1"
-    %define retrace_crontab_entry4 "#0 6,18 * * * /usr/bin/retrace-server-reposync fedora 16 x86_64 >> /var/log/retrace-server/reposync_error.log 2>&1"
-    %define retrace_crontab_entry5 "#0 8,20 * * * /usr/bin/retrace-server-reposync fedora rawhide i386 >> /var/log/retrace-server/reposync_error.log 2>&1"
-    %define retrace_crontab_entry6 "#0 10,22 * * * /usr/bin/retrace-server-reposync fedora rawhide x86_64 >> /var/log/retrace-server/reposync_error.log 2>&1"
+    %define retrace_crontab_entry1 "# 0 0,12 * * * /usr/bin/retrace-server-reposync fedora 15 i386 >> /var/log/retrace-server/reposync_error.log 2>&1"
+    %define retrace_crontab_entry2 "# 0 2,14 * * * /usr/bin/retrace-server-reposync fedora 15 x86_64 >> /var/log/retrace-server/reposync_error.log 2>&1"
+    %define retrace_crontab_entry3 "# 0 4,16 * * * /usr/bin/retrace-server-reposync fedora 16 i386 >> /var/log/retrace-server/reposync_error.log 2>&1"
+    %define retrace_crontab_entry4 "# 0 6,18 * * * /usr/bin/retrace-server-reposync fedora 16 x86_64 >> /var/log/retrace-server/reposync_error.log 2>&1"
+    %define retrace_crontab_entry5 "# 0 8,20 * * * /usr/bin/retrace-server-reposync fedora rawhide i386 >> /var/log/retrace-server/reposync_error.log 2>&1"
+    %define retrace_crontab_entry6 "# 0 10,22 * * * /usr/bin/retrace-server-reposync fedora rawhide x86_64 >> /var/log/retrace-server/reposync_error.log 2>&1"
 
     (crontab -u retrace -l 2> /dev/null; echo %{retrace_crontab_entry0}; \
      echo %{retrace_crontab_entry1}; echo %{retrace_crontab_entry2}; \
--- a/src/retrace/retrace.py
+++ b/src/retrace/retrace.py
@@ -887,7 +887,7 @@ class RetraceTask:
     MOCK_DEFAULT_CFG = "default.cfg"
     MOCK_SITE_DEFAULTS_CFG = "site-defaults.cfg"
     MOCK_LOGGING_INI = "logging.ini"
-    DOCKERFILE = "Dockerfile"
+    CONTAINERFILE = "Containerfile"
 
     def __init__(self, taskid: Optional[Union[int, str]] = None):
         """Creates a new task if taskid is None,
@@ -1879,7 +1879,7 @@ class RetraceTask:
                           RetraceTask.CRASHRC_FILE, RetraceTask.CRASH_CMD_FILE,
                           RetraceTask.URL_FILE, RetraceTask.MOCK_LOG_DIR,
                           RetraceTask.VMLINUX_FILE, RetraceTask.BUGZILLANO_FILE,
-                          RetraceTask.DOCKERFILE, RetraceTask.C2P_LOG_FILE]:
+                          RetraceTask.CONTAINERFILE, RetraceTask.C2P_LOG_FILE]:
                 continue
 
             try:
--- a/src/retrace/retrace.py
+++ b/src/retrace/retrace.py
@@ -13,7 +13,7 @@ import urllib.parse
 import urllib.request
 from pathlib import Path
 from signal import getsignal, signal, SIG_DFL, SIGPIPE
-from subprocess import PIPE, STDOUT, DEVNULL, TimeoutExpired, run
+from subprocess import DEVNULL, PIPE, STDOUT, TimeoutExpired, run
 from typing import Dict, List, Optional, Set, Tuple, Union
 import magic
 
@@ -306,7 +306,7 @@ def run_gdb(savedir: Union[str, Path], plugin, repopath: str, taskid: int):
         podman_build_call = [PODMAN_BIN, "build",
                              "--quiet",
                              "--force-rm",
-                             "--file", savedir / RetraceTask.DOCKERFILE,
+                             "--file", str(savedir / RetraceTask.DOCKERFILE),
                              "--volume=%s:%s:ro" % (repopath, repopath)]
 
         if CONFIG["RequireGPGCheck"]:
--- a/src/retrace/retrace.py
+++ b/src/retrace/retrace.py
@@ -1878,8 +1878,7 @@ class RetraceTask:
                           RetraceTask.TYPE_FILE, RetraceTask.RESULTS_DIR,
                           RetraceTask.CRASHRC_FILE, RetraceTask.CRASH_CMD_FILE,
                           RetraceTask.URL_FILE, RetraceTask.MOCK_LOG_DIR,
-                          RetraceTask.VMLINUX_FILE, RetraceTask.BUGZILLANO_FILE,
-                          RetraceTask.CONTAINERFILE, RetraceTask.C2P_LOG_FILE]:
+                          RetraceTask.VMLINUX_FILE, RetraceTask.BUGZILLANO_FILE]:
                 continue
 
             try:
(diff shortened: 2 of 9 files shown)


mgrabovsky:
@DaveWysochanskiRH Could you please verify these changes don't break anything for you?

DaveWysochanskiRH:
> @DaveWysochanskiRH Could you please verify these changes don't break anything for you?

Ok let me pull your patches into our dev environment.

DaveWysochanskiRH (commented):
Why do you need to change this?  These are exceptionally difficult to test, though this may be ok.

mgrabovsky:
Yes, it's a big change, but I feel like we need to change the way we do retracing in containers. We've had several difficulties with the Fedora retrace server instance with severe bugs popping up one after another in the past half-year or so. The approach in this PR simplifies the Podman workflow a tiny bit and makes it much easier to debug potential future failures.

DaveWysochanskiRH:
> Yes, it's a big change, but I feel like we need to change the way we do retracing in containers. We've had several difficulties with the Fedora retrace server instance with severe bugs popping up one after another in the past half-year or so. The approach in this PR simplifies the Podman workflow a tiny bit and makes it much easier to debug potential future failures. So it's a fair price for us to pay at the moment.

Sorry - let me be clear.  I was only looking at commit aace7dfeda089b98de8788f12314530291130f84 and wondering why that was lumped into this changeset since it's unrelated to podman.

DaveWysochanskiRH:
We don't use podman configuration option in our dev or prod machines unfortunately (we use RetraceEnvironment = native due to interactive mode and need for various tools such as pykdump, etc).  I'll see if I can configure the dev machine for it but right now I'm having some problems with it.

mgrabovsky:
> Sorry - let me be clear. I was only looking at commit [aace7df](https://github.com/abrt/retrace-server/commit/aace7dfeda089b98de8788f12314530291130f84) and wondering why that was lumped into this changeset since it's unrelated to podman.

Oh, right. That's just something that jumped at me from a Pylint scan. You're right it's not related to the Podman stuff, but the behaviour should be unchanged. It's just that `\d` is not a valid escape sequence in a Python string. (It should either be `\\d` or the `r` prefix should be used as I did here.)

> We don't use podman configuration option in our dev or prod machines unfortunately (we use RetraceEnvironment = native due to interactive mode and need for various tools such as pykdump, etc). I'll see if I can configure the dev machine for it but right now I'm having some problems with it.

I'm sorry if I wasn't clear. I don't need you to test this with the Podman backend, I'd just like to ask you to check that the code works fine in your usual use cases, even if you use the native backend. I was refactoring some functions and moving stuff around, so I just want to make sure nothing else broke in the process of fixing the Podman side.

DaveWysochanskiRH:
FYI, I configured podman in our dev environment and got the below.  NOTE: this does not look like a new problem, but maybe you can fix this here too, and I should be able to give further testing.  I would like to shift to podman if possible but it may take a bit of time.
~~~
[2021-04-20 11:58:53] [D] Parsing kernel version '4.18.0-94.el8.x86_64'
[2021-04-20 11:58:53] [D] Version: '4.18.0'; Release: '94.el8'; Arch: 'x86_64'; Flavour: 'None'; Realtime: False
[2021-04-20 11:58:53] [D] Determined kernel version: 4.18.0-94.el8.x86_64
[2021-04-20 11:58:53] [I] Preparing environment for backtrace generation
[2021-04-20 11:58:53] [E] Unable to read distribution and version from 'release' file: [Errno 2] No such file or directory: '/cores/retrace/tasks/860796081/crash/release'
[2021-04-20 11:58:53] [E] Cannot guess distribution and version without package
[2021-04-20 11:58:53] [E] Task failed: None
[2021-04-20 11:58:53] [D] Backtrace: None
  File "/usr/lib/python3.6/site-packages/retrace/retrace_worker.py", line 288, in read_release_file
    with release_path.open() as release_file:
  File "/usr/lib64/python3.6/pathlib.py", line 1183, in open
    opener=self._opener)
  File "/usr/lib64/python3.6/pathlib.py", line 1037, in _opener
    return self._accessor.open(self, flags, mode)
  File "/usr/lib64/python3.6/pathlib.py", line 387, in wrapped
    return strfunc(str(pathobj), *args)
FileNotFoundError: [Errno 2] No such file or directory: '/cores/retrace/tasks/860796081/crash/release'

During handling of the above exception, another exception occurred:

  File "/usr/lib/python3.6/site-packages/retrace/retrace_worker.py", line 1015, in start
    self.start_vmcore(custom_kernelver=kernelver)
  File "/usr/lib/python3.6/site-packages/retrace/retrace_worker.py", line 817, in start_vmcore
    release, distribution, version, _ = self.read_release_file(crashdir, None)
  File "/usr/lib/python3.6/site-packages/retrace/retrace_worker.py", line 307, in read_release_file
    self._fail()
  File "/usr/lib/python3.6/site-packages/retrace/retrace_worker.py", line 181, in _fail
    raise RetraceWorkerError(errorcode=errorcode)
retrace.retrace.RetraceWorkerError: None
~~~

DaveWysochanskiRH:
> > Yes, it's a big change, but I feel like we need to change the way we do retracing in containers. We've had several difficulties with the Fedora retrace server instance with severe bugs popping up one after another in the past half-year or so. The approach in this PR simplifies the Podman workflow a tiny bit and makes it much easier to debug potential future failures. So it's a fair price for us to pay at the moment.
> 
> Sorry - let me be clear. I was only looking at commit [aace7df](https://github.com/abrt/retrace-server/commit/aace7dfeda089b98de8788f12314530291130f84) and wondering why that was lumped into this changeset since it's unrelated to podman.

I'm going to see if I can create a small test program to ensure this type of change is well covered.  I'm not sure if you're open to maybe pulling this one out of this series, and maybe putting into a 'misc fixes' series a bit later?  That would give a bit more time for a test program and validate this type of change.  I certainly don't love this code but it's critical to proper retrace-server function with kernel vmcores.

DaveWysochanskiRH:
> > > Yes, it's a big change, but I feel like we need to change the way we do retracing in containers. We've had several difficulties with the Fedora retrace server instance with severe bugs popping up one after another in the past half-year or so. The approach in this PR simplifies the Podman workflow a tiny bit and makes it much easier to debug potential future failures. So it's a fair price for us to pay at the moment.
> > 
> > 
> > Sorry - let me be clear. I was only looking at commit [aace7df](https://github.com/abrt/retrace-server/commit/aace7dfeda089b98de8788f12314530291130f84) and wondering why that was lumped into this changeset since it's unrelated to podman.
> 
> I'm going to see if I can create a small test program to ensure this type of change is well covered. I'm not sure if you're open to maybe pulling this one out of this series, and maybe putting into a 'misc fixes' series a bit later? That would give a bit more time for a test program and validate this type of change. I certainly don't love this code but it's critical to proper retrace-server function with kernel vmcores.

Ok I wrote a small program and compared the kernelver generated with this new patchset with the existing "kernelver" file on all vmcores we have currently on our production machine.  The program spit out 3 vmcores with discrepancies and I'm investigating.  So please hold this patch for now or remove to another patchset.

mgrabovsky:
> Ok I wrote a small program and compared the kernelver generated with this new patchset with the existing "kernelver" file on all vmcores we have currently on our production machine. The program spit out 3 vmcores with discrepancies and I'm investigating. So please hold this patch for now or remove to another patchset.

Sorry about that. I see where the problem might have been and I've pushed a fixed version now. If it still doesn't work, I will this specific change and submit it in a separate PR, no big deal.

Regarding the crash above, I've just pushed a fix as well, although it appears that the Podman backend hasn't been tested with retracing vmcores, so there might a some more problems in the pipline. Fingers crossed.

DaveWysochanskiRH:
@mgrabovsky the latest patches got further with podman so thanks for fixing that, but looks like I have some permissions issues (this is RHEL8, FWIW):
~~~
[2021-04-21 08:39:58] [D] Parsing kernel version '4.18.0-94.el8.x86_64'
[2021-04-21 08:39:58] [D] Version: '4.18.0'; Release: '94.el8'; Arch: 'x86_64'; Flavour: 'None'; Realtime: False
[2021-04-21 08:39:58] [D] Determined kernel version: 4.18.0-94.el8.x86_64
[2021-04-21 08:39:58] [I] Preparing environment for backtrace generation
[2021-04-21 08:39:59] [E] Task failed: Unable to build podman container: Error: error creating runtime static files directory: mkdir /usr/share/retrace-server/.local: permission denied

[2021-04-21 08:39:59] [D] Backtrace: Unable to build podman container: Error: error creating runtime static files directory: mkdir /usr/share/retrace-server/.local: permission denied
  File "/usr/lib/python3.6/site-packages/retrace/retrace_worker.py", line 1114, in start
    self.start_vmcore(custom_kernelver=kernelver)
  File "/usr/lib/python3.6/site-packages/retrace/retrace_worker.py", line 944, in start_vmcore
    raise Exception(f"Unable to build podman container: {child.stdout}")
Exception: Unable to build podman container: Error: error creating runtime static files directory: mkdir /usr/share/retrace-server/.local: permission denied
~~~

DaveWysochanskiRH:
> > Ok I wrote a small program and compared the kernelver generated with this new patchset with the existing "kernelver" file on all vmcores we have currently on our production machine. The program spit out 3 vmcores with discrepancies and I'm investigating. So please hold this patch for now or remove to another patchset.
> 
> Sorry about that. I see where the problem might have been and I've pushed a fixed version now. If it still doesn't work, I will this specific change and submit it in a separate PR, no big deal.
> 
I think actually your original change may be ok, but if you're ok with it, can you pull out separately so I have more time to confirm?  I'd like to do more testing with say 2 weeks worth of production vmcores.  That will ensure it's well tested on a wide range of vmcores.

mgrabovsky:
OK, I've pulled that commit for now.

That failure looks like Podman thinks `/usr/share/retrace-server/` is retrace user's home directory. Could you please check that the home directory is set correctly, preferably something like `/home/retrace` with the correct permissions?

DaveWysochanskiRH:
> OK, I've pulled that commit for now.
> 
> That failure looks like Podman thinks `/usr/share/retrace-server/` is retrace user's home directory. Could you please check that the home directory is set correctly, preferably something like `/home/retrace` with the correct permissions?

Well unfortunately I think "/usr/share/retrace-server" is hardcoded into various places (see src/manager.wsgi for example) and this is a result of the install.  This is retrace user's home dir and it's owned by root, so maybe we need some separate patches here.

DaveWysochanskiRH:
> > OK, I've pulled that commit for now.
> > That failure looks like Podman thinks `/usr/share/retrace-server/` is retrace user's home directory. Could you please check that the home directory is set correctly, preferably something like `/home/retrace` with the correct permissions?
> 
> Well unfortunately I think "/usr/share/retrace-server" is hardcoded into various places (see src/manager.wsgi for example) and this is a result of the install. This is retrace user's home dir and it's owned by root, so maybe we need some separate patches here.

Ok so I uninstalled retrace-server, removed 'retrace' user, then reinstalled, got further.  Need to setup a few things (from DEPLOYING.md) - now having problems with container registries.

mgrabovsky:
> Well unfortunately I think "/usr/share/retrace-server" is hardcoded into various places (see src/manager.wsgi for example) and this is a result of the install. This is retrace user's home dir and it's owned by root, so maybe we need some separate patches here.

I'm honestly not sure why these references are in the code, but I'm sure they're unrelated to this problem. The retrace user's home directory should be `/var/lib/retrace` by default in recent versions. (I wouldn't be surprised if it used to be `/usr/share/retrace-server`, though I hope that wasn't the case.)

DaveWysochanskiRH:
> > Well unfortunately I think "/usr/share/retrace-server" is hardcoded into various places (see src/manager.wsgi for example) and this is a result of the install. This is retrace user's home dir and it's owned by root, so maybe we need some separate patches here.
> 
> I'm honestly not sure why these references are in the code, but I'm sure they're unrelated to this problem. The retrace user's home directory should be `/var/lib/retrace` by default in recent versions. (I wouldn't be surprised if it used to be `/usr/share/retrace-server`, though I hope that wasn't the case.)

That must have been the case because both my dev and prod instances are very old and I've never manually edited retrace user's home dir.  Also I never actually tested 'podman' mode until this patchset so it probably went unnoticed.  It is a fairly easy fixup though maybe another patch to consider is to check for the old path during a upgrade/install and fix it up.

mgrabovsky:
@michalfabik @msrb Please review.

We've agreed with Dave to keep the vmcores stuff for later. I'll file issues for them and take a look at them in two weeks.

DaveWysochanskiRH (commented):
This looks correct, but from what I can tell it doesn't work.  I suspect it's due to the fact that on an upgrade, retrace-server will be running and you'll get an "in use" error so usermod will fail.  I tried manually and this is what I got.  I'm not sure if you were attempting to only address clean install, or both install and upgrade.

# /usr/sbin/usermod -d /var/lib/retrace retrace
usermod: user retrace is currently used by process 939762

DaveWysochanskiRH:
I tested all these patches in "native" mode and they look good.  The only thing I found that still does not work is the commit 14c63bf that should upgrade the home dir for retrace user.

DaveWysochanskiRH:
NOTE: As a POC patch, I confirmed stopping httpd before the usermod works:
~~~
diff --git a/retrace-server.spec b/retrace-server.spec
index 50b1ec444441..10a42d3abb56 100644
--- a/retrace-server.spec
+++ b/retrace-server.spec
@@ -141,7 +141,9 @@ getent passwd retrace > /dev/null || useradd --system -g retrace -u %{retrace_gi
 # if retrace's home is %{_datadir}/%{name} change it to %{_sharedstatedir}/retrace
 if [ "$(getent passwd retrace | cut -d: -f6)" = "%{_datadir}/%{name}" ]
 then
-    /usr/sbin/usermod -d %{retrace_home} retrace 2> /dev/null || :
+    systemctl stop httpd
+    /usr/sbin/usermod -d %{retrace_home} retrace
+    systemctl start httpd
 fi
 exit 0
~~~

msrb:
I've been thinking about this and I think the solution for the home dir proposed by @DaveWysochanskiRH is the best we can do here.

I looked for some nice systemd macros that we could use here, but I haven't find any: https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/#_systemd

I'd maybe explicitly `Requires(pre): /usr/bin/systemctl` just to make sure that `systemctl` is on the system when we run the `%pre` section, but I cannot really imagine system without systemctl installed on it. So we should be fine either way, IMO.

+1 for the change proposed by Dave.

msrb:
Nah... now I realized that we don't actually start `httpd` when people install retrace-server package on their system, so we should probably check if the service is already running before stopping and starting it again.

src/retrace/retrace.py
```diff
@@ -295,63 +385,54 @@ def run_gdb(savedir: Union[str, Path], plugin, repopath: str, taskid: int):
...
-            podman_build_call.append("--volume=%s:%s:ro" % (RETRACE_GPG_KEYS, RETRACE_GPG_KEYS))
+        backend = LocalPodmanBackend(retrace_config=CONFIG)
 
-        if CONFIG["UseFafPackages"]:
-            faf_link_dir = CONFIG["FafLinkDir"]
-            log_debug("Using FAF repository")
-            podman_build_call.append("--volume=%s:%s" % (faf_link_dir, faf_link_dir))
+        # Start container from prepared release-specific image.
+        with backend.start_container(image_tag, taskid, repopath) as container:
+            # Copy coredump from crash directory to container.
```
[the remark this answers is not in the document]
mgrabovsky: This is nothing new, we used to copy the coredump in the previous implementation as well. Since the container is ephemeral, it shouldn't pose any long-term issues when it comes to storage, and it should be reasonably fast even for larger coredumps.

If need be, I suppose we could mount the directory where the coredump resides as a volume and avoid the copy. However, that seems to me to add unnecessary complexity at the moment.

retrace-server.spec
```diff
@@ -147,12 +155,12 @@ if [ "$1" = 1 ]
...
-    %define retrace_crontab_entry1 "#0 0,12 * * * /usr/bin/retrace-server-reposync fedora 15 i386 >> /var/log/retrace-server/reposync_error.log 2>&1"
-    %define retrace_crontab_entry2 "#0 2,14 * * * /usr/bin/retrace-server-reposync fedora 15 x86_64 >> /var/log/retrace-server/reposync_error.log 2>&1"
-    %define retrace_crontab_entry3 "#0 4,16 * * * /usr/bin/retrace-server-reposync fedora 16 i386 >> /var/log/retrace-server/reposync_error.log 2>&1"
-    %define retrace_crontab_entry4 "#0 6,18 * * * /usr/bin/retrace-server-reposync fedora 16 x86_64 >> /var/log/retrace-server/reposync_error.log 2>&1"
-    %define retrace_crontab_entry5 "#0 8,20 * * * /usr/bin/retrace-server-reposync fedora rawhide i386 >> /var/log/retrace-server/reposync_error.log 2>&1"
-    %define retrace_crontab_entry6 "#0 10,22 * * * /usr/bin/retrace-server-reposync fedora rawhide x86_64 >> /var/log/retrace-server/reposync_error.log 2>&1"
+    %define retrace_crontab_entry1 "# 0 0,12 * * * /usr/bin/retrace-server-reposync fedora 15 i386 >> /var/log/retrace-server/reposync_error.log 2>&1"
```
msrb, line 158: `retrace-server-reposync fedora 15`

We could probably update the default versions here. But that is unrelated to this PR and we can do it later.

mgrabovsky: Yeah, this part is funny, though ideally I'd like to find a solution that doesn't rely on us updating this every 6 months.

mgrabovsky:
@msrb Thanks for your research. I've just updated the PR around the scriptlet. Please take a look and confirm that it's fine this way.

mgrabovsky:
Just realised the latest change pretty much rules out running Retrace Server inside a container now since they usually don't come with systemd.

msrb (approved):
LGTM

msrb:
> Just realised the latest change pretty much rules out running Retrace Server inside a container now since they usually don't come with systemd.

This should be fine. I think we wouldn't be installing retrace server to the container image from RPMs anyway.

mgrabovsky:
OK, let's merge this and see what breaks in production 👷

DaveWysochanskiRH:
FYI, I tested the final patches in dev and now deployed to production.

PR #429, opened 2021-06-16
Remove old test, create new one

mgrabovsky:
Remove an old attempt at an integration test, which didn't work reliably anyway. Create a few new unit tests for the Podman backend class.

Moreover, fix a bug found by the test. An off-by-one error made the short container ID one character too short.


mgrabovsky:
There's something wrong with the tests in the EPEL 8 build, so let's keep this open for now. I'll get back to it later.

msrb:
I was able to run the tests successfully on my Fedora 34 workstation (outside of RPM). I can try on RHEL 8 (+EPEL) as well.

msrb:
So far, it's +1 from me. Let's test that EPEL further.

mgrabovsky:
@msrb I've fixed the tests on EPEL (more precisely, for Python 3.8). Could you please provide your green check mark? :)

msrb (approved):
LGTM :wink: Good catch with the older Python in RHEL 8.

PR #432, opened 2021-07-01
Fix bare `except` occurrences

mgrabovsky:
A bare `except:` statement catches all exception inheriting from `BareException` including `SystemExit` and `KeyboardInterrupt`. That is rarely what we want and whenever we do, we should be explicit about that and handle each case separately.

Problem pointed out by LGTM.


michalfabik (dismissed):
I mean, sure, it would be nice to have something more descriptive than just Exception but I suppose there's no need to bother in a case like this.

PR #433, opened 2021-07-17
Add debuginfod support to retrace-server

aaronmerey:
Debuginfod is an HTTP server for distributing ELF, DWARF and source
code.

This patch allows retrace-server to skip aquiring packages needed to
generate a backtrace from a corefile. Instead, when configured to do so,
GDB's built-in debuginfod client will automatically query debuginfod
servers in order to aquire the necessary debugging resources during
backtrace generation.

Two new options are added to retrace-server.conf: 'DebuginfodEnable'
determines whether the server will permit debuginfod functionality and
'DebuginfodURLs' is a list of servers to be queried by the GDB
debuginfod client.

'--debuginfod' options are also added to retrace-server-{interact, task,
worker}.

Documentation is updated to mention debuginfod.

For more information regarding debuginfod, see
https://sourceware.org/elfutils/Debuginfod.html

Note that upstream improvements to GDB's debuginfod corefile support
are currently in the works. Until these improvements are available
in the Fedora builds of GDB available from DNF, it is recommended
that retrace-server uses the following GDB rpms for the best
debuginfod performance when testing or experimenting with this PR:
https://copr.fedorainfracloud.org/coprs/amerey/gdb-debuginfod-core/

### DIFF
--- a/doc/retrace-server.texi
+++ b/doc/retrace-server.texi
@@ -427,6 +427,15 @@ The files containing the debugging symbols for the binary and libraries
 (build-ids are used to find the pairs) need to be available to GDB.
 @end itemize
 
+Alternatively if GDB is built with @command{debuginfod} support and
+the @var{DebuginfodEnable} option is enabled in the Retrace Server's
+configuration, then GDB will query the @command{debuginfod} servers
+listed in @var{DebuginfodURLs} in the configuration file for all
+of the debugging resources needed to generate the backtrace. The
+task also needs to be created with the @var{X-Debuginfod} header.
+For more information regarding @command{debuginfod} see
+@uref{https://sourceware.org/elfutils/Debuginfod.html}.
+
 The chroot environments are created and managed by @command{mock}, and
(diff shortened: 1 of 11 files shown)


mgrabovsky:
Thanks for the patches, Aaron. It's nice to have a working implementation to play around with at last.

The code looks OK overall, though I have two remarks:

* We would prefer that debuginfod be invisible to the client when retracing remotely, i.e. no special headers and no changes on the abrt-retrace-client side. A global switch in `retrace-server.conf` on the server side should be enough.
* This PR adds debuginfod support to the Mock backend only. However, we've moved on to the Podman backend on the Fedora instance, so we need support for that as well. This would entail some more changes at least in in `run_gdb()` and `RetraceWorker.ensure_image_exists()`.

mgrabovsky:
We'll be integrating debuginfod in a slightly different way, so I'm closing this as well as abrt/abt#1574. Thanks for the contribution, though, it looks like debuginfod is really easy to integrate and work with.

PR #438, opened 2021-10-17
manager: Disallow NULL string in custom core location on manager page

DaveWysochanskiRH:
Before this patch, on the manager page if you accidentally click
"Create vmcore task" without typing anything into the "Custom
core location" box a task will be created and immediately fail
with a message similar to the following:
[2021-10-11 22:39:11] [I] Downloading remote resources
[2021-10-11 22:39:11] [I] Retrieving remote file ''
[2021-10-11 22:39:11] [W] malformed URL
[2021-10-11 22:39:11] [W] No files found in the tarball
[2021-10-11 22:39:11] [I] Analyzing crash data
[2021-10-11 22:39:11] [E] Task failed: Crash directory does not contain required file 'vmcore'
[2021-10-11 22:39:11] [D] Backtrace: Crash directory does not contain required file 'vmcore'
  File "/usr/lib/python3.6/site-packages/retrace/retrace_worker.py", line 1123, in start
    raise Exception("Crash directory does not contain required file '%s'" % required_file)
Exception: Crash directory does not contain required file 'vmcore'

This patch fixes the above by checking for a 0 length string in
the custom_url, and returns a 404 error with the message
"URL cannot be NULL string", rather than creating a task which
always fails.


DaveWysochanskiRH:
FYI, this was actually happening in our production environment which is why I submitted this patch.

There has been another problem that has puzzled me for a while - multiple tasks may get stuck in "Available" listing on the manager page with no file or any status - so they are very early but get stuck there and nothing happens.  They also do not get cleaned up.  I looked at the timestamp on the directories of these tasks and found they were at the same minute as ones that were failed with this "NULL custom URL" error, so I think those hung "available" tasks may be related to this problem as well.

PR #448, opened 2021-12-05
manager: Add permissions check to custom_url to avoid failed tasks

DaveWysochanskiRH:
Tasks that are created with custom_url pointing at local files
with incorrect permissions will almost immediately fail, leaving
failed tasks and delayed feedback to the user.
Move the permissions check into the __custom__ URL parsing inside
the manager.wsgi file, which is much sooner than where it is
currently done inside download_remote().  Checking permissions
early gives the user early feedback via the http interface,
and we also avoid creating tasks that will immediately fail
due to improper permissions on the file.

### DIFF
--- a/src/manager.wsgi
+++ b/src/manager.wsgi
@@ -3,6 +3,7 @@ import fnmatch
 import re
 import time
 import urllib
+import stat
 from webob import Request
 from pathlib import Path
 from typing import Any, Dict, Optional, List
@@ -485,8 +486,23 @@ def application(environ, start_response):
     elif filename and filename == "__custom__":
         POST = request.POST
 
-        if "custom_url" in POST and len(POST["custom_url"]) == 0:
-            return response(start_response, "400 Bad Request", _("URL cannot be NULL string"))
+        if "custom_url" in POST:
+            url = POST["custom_url"]
+            if not url:
+                return response(start_response, "400 Bad Request", _("URL cannot be NULL string"))
+            if url.startswith("/") or url.startswith("file:///"):
+                if url.startswith("file://"):
+                    path = Path(urllib.parse.unquote(url[7:]))
+                else:
+                    path = Path(url)
+
+                # Just check the file exists and we can open it for binary read
+                # Read the file later in RetraceWorker:start_vmcore()
+                try:
+                    with open(path, "rb"):
+                        pass
+                except Exception as e:
+                    return response(start_response, "400 Bad Request", _("Cannot open URL for reading: '%s'" % e))
 
         qs_base = []
         if "md5sum" in POST and POST["md5sum"] == "on":


DaveWysochanskiRH:
One optimization could be to also check the "custom_vmem_url" path in the same manner.  In practice, vmem files don't show up too often, and it would be a corner of a corner case for the "custom_url" path to be readable and yet the "custom_vmem_url" to be unreadable, so I didn't include it.

DaveWysochanskiRH:
Still testing... found one problem with non-existent files.

DaveWysochanskiRH:
Ok I think 8edcbe0 fixes this and does not introduce any new issues.  Main tests were:
1. Check for file that exists but is not readable (permission)
2. Check for file that does not exist 
The first scenario seems fairly common while the second scenario does happen it is not nearly as often.

src/manager.wsgi
```diff
@@ -485,8 +486,22 @@ def application(environ, start_response):
     elif filename and filename == "__custom__":
         POST = request.POST
 
-        if "custom_url" in POST and len(POST["custom_url"]) == 0:
-            return response(start_response, "400 Bad Request", _("URL cannot be NULL string"))
+        if "custom_url" in POST:
+            url = POST["custom_url"]
+            if len(url) == 0:
```
mgrabovsky, line 491: I suggest making this more Pythonic.

```suggestion
            if not url:
```

src/manager.wsgi
```diff
@@ -485,8 +486,22 @@ def application(environ, start_response):
...
+            if url.startswith("/") or url.startswith("file:///"):
+                if url.startswith("file://"):
+                    path = Path(urllib.parse.unquote(url[7:]))
+                else:
+                    path = Path(url)
+
+                try:
+                    st = path.stat()
+                    with open(path, "rb") as fd:
+                        fd.seek(0)
```
mgrabovsky, line 502: What is the purpose of this block of code? Could you please add a brief comment explaining what it does and why it's there?

src/manager.wsgi
DaveWysochanskiRH: sure, will do!

src/manager.wsgi
DaveWysochanskiRH: Purpose is to check for existence of the file as well as permissions.
If the file does not exist the response is:
URL '/path/vmcore-nonexist' cannot open for reading error '[Errno 2] No such file or directory: '/path/vmcore-nonexist''

If the file exists but permissions are wrong the response is:
URL '/path/vmcore-noread' cannot open for reading error '[Errno 13] Permission denied: '/path/vmcore-noread''

The response message above could be improved - don't need the path twice for example.

Also, the 'stat' is unnecessary - trying to open for read is enough, so let me see if I can simplify this and fix it up.

DaveWysochanskiRH: Something like this is probably more pythonic, though not sure it does exactly what is needed:
~~~
@@ -496,12 +496,11 @@ def application(environ, start_response):
                 else:
                     path = Path(url)
 
-                try:
-                    st = path.stat()
-                    with open(path, "rb") as fd:
-                        fd.seek(0)
-                except Exception as e:
-                    return response(start_response, "400 Bad Request", _("URL '%s' cannot open for reading error '%s'" % (url, e)))
+               if not path.exists():
+                    return response(start_response, "400 Bad Request", _("URL '%s' does not exist" % url))
+
+                if not path.is_file():
+                    return response(start_response, "400 Bad Request", _("URL '%s' is not readable" % url))
 
         qs_base = []
         if "md5sum" in POST and POST["md5sum"] == "on":
~~~

DaveWysochanskiRH: While the above was more pythonic, the is_file() check didn't work to check whether the file could be opened read-only.
After some more looking, I felt the explicit open around a try .. except was the best check.  Feel free to suggest an alternative if you don't like that and we can achieve the equivalent.  Another possibility I found was `access(file, os.R_OK)` but that just gives a true/false and I think the errno string of the try..except is better.

DaveWysochanskiRH:
In 40f1f620f7aeb43bb485c4d67557aac6eeaeda99 I also updated the commit message as well as the code, to hopefully make this all more clear.

DaveWysochanskiRH:
Also, 40f1f62 also catches other problems seen in production, such as someone giving a directory instead of a file.  In that case, the following response is given:
Cannot open URL for reading: '[Errno 21] Is a directory: '/foo/bar''

DaveWysochanskiRH:
@mgrabovsky any further comments / suggestions or can this be merged?  For testing, last week I put 40f1f62 into our system and asked for feedback but so far no feedback and it seems to fix the problems with failed tasks to to perms or non-existent files.

mgrabovsky:
Sorry, I haven't got around to re-reviewing this yet. I'll get to it to tomorrow.

src/manager.wsgi
```diff
@@ -485,8 +486,21 @@ def application(environ, start_response):
...
+                return response(start_response, "400 Bad Request", _("URL cannot be NULL string"))
+            if url.startswith("/") or url.startswith("file:///"):
+                if url.startswith("file://"):
+                    path = Path(urllib.parse.unquote(url[7:]))
+                else:
+                    path = Path(url)
+
+                # Check the file exists and we can open it for binary read
+                try:
+                    open(path, "rb")
```
mgrabovsky, line 501: Looks fine overall. Only one nitpick: The way this line is written now is basically a resource leak. I suggest wrapping it in a dummy `with`, such as

```suggestion
                    with open(path, "rb"):
                        # We only check if the file can be opened, we don't read it yet.
                        pass
```

DaveWysochanskiRH: Hmmm, I thought the try/except would insulate the leak but maybe I'm wrong there?  I can fix this up.

mgrabovsky (approved):
Looks good, thanks!

PR #450, opened 2021-12-19
bugzilla-query: Use 'limit' option and search only recently modified …

DaveWysochanskiRH:
…bugs

Set the 'limit' option to bzapi.build_query() function to ensure we
are not artificially limited to a small default number of bugs.
Note the 'limit' option is new in python-bugzilla 3.1.0.
In production we were seeing an artificial limit of 20 bugs get
queried without setting this option.  If we ever meet our defined
limit of 1000 bugs in the query this means we probably did not
query all bugs.

In addition, cut down on the number of bugs in the query by only
querying those with recent modification times.  Here we pick 7 days,
somewhat arbitrarily, but with the understanding that the cronjob
that runs bugzilla-query should be run at least once per week.  We
can look at only the recently modified bugs because we are looking
for additions of retrace-server task strings in the bug comment
text.

### DIFF
--- a/src/retrace-server-bugzilla-query
+++ b/src/retrace-server-bugzilla-query
@@ -1,7 +1,8 @@
 #!/usr/bin/python3
 """
-Query bugzilla for component bugs and search through comments for trigger words. Output results
-to a log and for tasks found in the comments set bugzillano field with bugzilla ids.
+Query bugzilla for component bugs and search through comments for
+trigger words.  Output results to a log and for tasks found in the
+comments set bugzillano field with bugzilla ids.
 """
 
 import sys
@@ -32,7 +33,8 @@ if __name__ == "__main__":
         bz_creds = CONFIG["BugzillaCredentials"]
 
         try:
-            bzapi = bugzilla.Bugzilla(url=bz_url, cookiefile=None, tokenfile=None)
+            bzapi = bugzilla.Bugzilla(url=bz_url, cookiefile=None,
+                                      tokenfile=None)
         except (bugzilla.BugzillaError, ValueError) as e:
             log.write("Exception: {0}".format(e))
             sys.exit(1)
@@ -81,7 +83,7 @@ if __name__ == "__main__":
                 found_ids: Set[str] = set()
                 for comment in bug.comments:
                     for i, trigger_word in enumerate(trigger_words):
-                        # Only if trigger word is in comment do we use slower regex
+                        # Use slower regex only if trigger word is in comment
                         if trigger_word in comment["text"]:
                             m = re.findall(regexes[i], comment["text"])
                             found_ids.update(m)
@@ -105,10 +107,13 @@ if __name__ == "__main__":
 
             existing_tasks.append(taskdir.name)
 
-        log.write("------------Existing tasks with found bugzillas-----------\n")
+        log.write("------------"
+                  "Existing tasks with found bugzillas"
+                  "------------\n")
         for taskid in found_tasks:
             if taskid in existing_tasks:
-                log.write("Task {0} has following bugzilla(s): {1}.\n".format(taskid, ", ".join(found_tasks[taskid])))
+                log.write("Task {0} has following bugzilla(s): {1}."
+                          "\n".format(taskid, ", ".join(found_tasks[taskid])))
                 try:
                     task = RetraceTask(taskid)
                 except Exception:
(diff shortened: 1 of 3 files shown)


src/retrace-server-bugzilla-query
```diff
@@ -7,6 +7,7 @@ to a log and for tasks found in the comments set bugzillano field with bugzilla
 import sys
 import re
 import time
+from datetime import datetime, timedelta
```
mgrabovsky, line 11: Drop this empty line please.

mgrabovsky (commented):
Do you think it would be useful to have both of the new values (`BUGZILLA_QUERY_LIMIT` and the 7 days in `query["last_change_time"]`) in the config? The values you propose look reasonable, I'm just thinking it would be easier to, for instance, adjust the time horizon to 30 days, if needed, in the config than in code.

Also, please shorten the first line of your commit message to about 50 characters, otherwise the logs in Git look a tad funny.

src/retrace-server-bugzilla-query
DaveWysochanskiRH: Yes I thought about making those parameters and I think you're right.  These settings are related to the DeleteTaskAfter and DeleteFailedTaskAfter too, as well as the cronjob settings for the various retrace-server-* tools.  So it's a bit of a complicated setup and probably they should all be configurable.

I also thought about the WARNING in there - probably that's not ideal.  It would be better to detect the limit is hit then run another query, assuming the API has a way to pickup where you left off - maybe start with a specific bug# and only ask for bugs > bug#?  I'll need to look a bit more at the bugzilla API.

DaveWysochanskiRH:
Ok, in 53cb39e I've fixed up the query to use "offset" as well as "limit" in the bugzilla API query, which is the way to query until all bugs are returned.  I moved things around a bit and hopefully it is all more readable.  As far as I can tell this does exactly what we want, and it should be safe for any future changes on the server end to limit the number of bugs returned.

A couple comments:
1. This removes the intermediate query to bugzilla which only was querying the bug_ids.  It's not clear if this was necessary before due to maybe some API limitation, but it's not necessary anymore.
2. Query with increasing offsets until we get 0 bugs back from the query.  The "while True" may look a bit odd but what this type of loop calls for is a "do .. while" and as far as I know this does not exist in Python.  The paranoia in me wondered about the possibility of an infinite loop here, and I almost put in a second "break" out of the loop with some "failsafe" code to count the iterations through the loop and break.  However, I think probably that is overkill and I didn't think it would see positive review so I left it out
3. I added BugzillaQueryLastChangeDelta which represents the limit on bz modified times for the bugzilla query, but left the "limit" number hardcoded at 500, as I couldn't see how the "limit" number adds value to be made a parameter.  In fact, we could probably omit the limit but then the internal limit of 20 would come into play and there would be many queries.
4. I reformatted a number of lines towards the end to fit in 79 columns and conform to the python [PEP-0008](https://www.python.org/dev/peps/pep-0008/)

DaveWysochanskiRH:
I decided to split out the reformatting changes into a separate commit on top of the main commit which implements the functional changes of this PR.  I am not sure how important cleanups like the line lengths are so if they are not important we can just drop that without affecting the functional changes.

src/retrace-server-bugzilla-query
```diff
@@ -47,37 +48,42 @@ if __name__ == "__main__":
...
-
-        # I don't want to run regex on each and every comment, at first try fast check
-        # if trigger word is in comment and if is, then use slower regex
+        config_status = CONFIG.get_list("BugzillaStatus")
         trigger_words = CONFIG.get_list("BugzillaTriggerWords")
         regexes = CONFIG.get_list("BugzillaRegExes")
+        if CONFIG["BugzillaQueryLastChangeDelta"]:
+            delta = CONFIG["BugzillaQueryLastChangeDelta"]
+        else:
+            delta = 7*24
```
mgrabovsky, line 62: You don't need this check because of the way the `Config` class works. Just `CONFIG["BugzillaQueryLastChangeDelta"]` should be enough and it also keeps the defaults where they should be.

src/retrace-server-bugzilla-query
```diff
@@ -47,37 +48,42 @@ if __name__ == "__main__":
...
-            cur_bugs = bzapi.query(query)
-            bugs = bugs[250:]
-            for bug in cur_bugs:
+        query["last_change_time"] = datetime.today() - timedelta(hours = delta)
+        bugzilla_offset = 0
+        bugzilla_bug_count = 0
+        while True:
+            bugzilla_offset += bugzilla_bug_count
+            query["offset"] = bugzilla_offset
+            query["limit"] = 500
```
mgrabovsky, line 74: I agree we don't need a config variable for the limit, but could you please pull the 500 out into a named constant at the top of the file? Just so we don't introduced any new magical numbers.

mgrabovsky (commented):
Looks pretty good now. I've only got two nitpicks.

src/retrace-server-bugzilla-query
DaveWysochanskiRH: Ok

src/retrace-server-bugzilla-query
DaveWysochanskiRH: Yes

DaveWysochanskiRH:
The two remaining items have been addressed in c09a689.  Thank you for the review - made the code much better.

Issue #445, opened 2021-11-08
Ensure makedumpfile savings are worthwhile and make a decision whether to keep stripped file


DaveWysochanskiRH:
Lately I've been noticing makedumpfile running on large vmcores (hundreds of GB), possibly taking a few hours, and ends up saving only a couple MB.  The end result is effectively we have double the space used, since the newly stripped vmcore is approx the same size as the original, and this new file defeats our built-in deduplication algorithm (inode based).

I doubt it's possible ahead of time to know how long makedumpfile will run, or what amount of space it will save.  But we should be able to add some sort of heuristic, so that the stripped file is kept only if it saves a certain significant %age of the original file size (maybe 10%?).

PR #452, opened 2022-01-18
Add logic to determine if we save a stripped vmcore

audrabaker:
Currently we do not check to see if running makedumpfile and creating a stripped vmcore is actually worthwhile. In some situations we may see little to no benefit to stripping down a vmcore and replacing the original with the newly stripped vmcore. For those situation, this patch adds the config variable SpaceSavePercent which is meant to indicate the percentage of space a stripped vmcore will need to save for us to choose to save it in place of the original. The proposed default is 10%. This patch also adjusts the logic so that the relevant message about makedumpfile will be printed (i.e. if makedumpfile is not run or we don't use the stripped vmcore from makedumpfile we do not print how much space we saved).

Signed-off-by: Audra Mitchell <aubaker@redhat.com> #445

### DIFF
--- a/src/retrace/retrace.py
+++ b/src/retrace/retrace.py
@@ -2105,6 +2105,8 @@ class KernelVMcore:
         if self._vmlinux is None:
             log_error("Cannot strip pages if vmlinux is not known for vmcore")
             return
+        start = time.time()
+        oldsize = self.get_path().stat().st_size
 
         newvmcore: Path = Path("%s.stripped" % self._vmcore_path)
         cmd = ["makedumpfile", "-c", "-d", "%d" % CONFIG["VmcoreDumpLevel"],
@@ -2122,8 +2124,25 @@ class KernelVMcore:
             log_warn("makedumpfile exited with %d" % child.returncode)
             if newvmcore.is_file():
                 newvmcore.unlink()
+            return 
+
+        #Calculate the % of storage space saved with the new makedumpfile core created. 
+        elif ((oldsize - newvmcore.stat().st_size) / int(oldsize) * 100 < CONFIG["VmcoreDumpSavePercent"]):
+            log_warn("makedumpfile did not strip enough pages (needed to save %{}). 
+                    Not saving new vmcore.".format(CONFIG["VmcoreDumpSavePercent"]))
+            format_number = "{:.2f}".format((int(oldsize) - int(newvmcore.stat().st_size)) / int(oldsize) * 100)
+            log_info("Percentage of space that would have been saved if we kept the stripped vmcore 
+                    is: %s" % format_number)
         else:
+            # makedumpfile did strip enough pages so save the new core
             newvmcore.rename(self._vmcore_path)
+            dur = int(time.time() - start)
+            newsize = self.get_path().stat().st_size
+            log_info("Stripped size: %s" % human_readable_size(newsize))
+            log_info("Makedumpfile took %d seconds and saved %s"
+                     % (dur, human_readable_size(oldsize - newsize)))
+        return
+
 
     #
     # In real-world testing, approximately 60% of the time the kernel
--- a/src/retrace/retrace.py
+++ b/src/retrace/retrace.py
@@ -2125,14 +2125,16 @@ class KernelVMcore:
             if newvmcore.is_file():
                 newvmcore.unlink()
             return 
-
-        #Calculate the % of storage space saved with the new makedumpfile core created. 
-        elif ((oldsize - newvmcore.stat().st_size) / int(oldsize) * 100 < CONFIG["VmcoreDumpSavePercent"]):
-            log_warn("makedumpfile did not strip enough pages (needed to save %{}). 
-                    Not saving new vmcore.".format(CONFIG["VmcoreDumpSavePercent"]))
-            format_number = "{:.2f}".format((int(oldsize) - int(newvmcore.stat().st_size)) / int(oldsize) * 100)
-            log_info("Percentage of space that would have been saved if we kept the stripped vmcore 
-                    is: %s" % format_number)
+        # Calculate the % of storage space saved with the new makedumpfile core created
+        if ((oldsize - newvmcore.stat().st_size) / oldsize * 100  
+            < CONFIG["VmcoreDumpSavePercent"]):
+            duration = int(time.time() - start)
+            log_warn("makedumpfile did not strip enough pages (needed to save at least "
+                     "{}%). Not saving new vmcore.".format(CONFIG["VmcoreDumpSavePercent"]))
+            saved_pct = (oldsize - newvmcore.stat().st_size) / oldsize * 100
+            log_info("makedumpfile took {} seconds. Would have saved {:.2f}% (or {}) "
+                     "if we kept the stripped vmcore.".format(duration, saved_pct,
+                     human_readable_size(oldsize - newvmcore.stat().st_size)))
         else:
             # makedumpfile did strip enough pages so save the new core
             newvmcore.rename(self._vmcore_path)
@@ -2141,7 +2143,6 @@ class KernelVMcore:
             log_info("Stripped size: %s" % human_readable_size(newsize))
             log_info("Makedumpfile took %d seconds and saved %s"
                      % (dur, human_readable_size(oldsize - newsize)))
-        return
 
 
     #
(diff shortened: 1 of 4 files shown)


DaveWysochanskiRH:
@audrabaker some feedback.

1. commit message needs limited to some reasonable # of columns (70?).  Do a 'git show' on other commits and you'll get an idea.  It shows as one very long line.
~~~
$ git show audrabaker/master
commit f1e647b28389b793ce54425d306e0dda150786af
Author: Audra Mitchell <aubaker@redhat.com>
Date:   Tue Jan 18 15:43:26 2022 -0500

    Add logic to determine if we save a stripped vmcore
    
    Currently we do not check to see if running makedumpfile and creating a stripped vmcore is actually worthwhile. In some situations we may see little to no benefit to stripping down a vmcore and 
    
    Signed-off-by: Audra Mitchell <aubaker@redhat.com>
~~~

2. I think the name of SpaceSavePercent is ok but probably needs improved.  We probably need vmcore in there because this only applies to vmcores.  Also, it's makedumpfile related so I wonder if we can clue in there too.  The only other parameter though is "VmcoreDumpLevel" so maybe "VmcoreDumpSavePercent" or is a little long but at least it's grouped somewhat with the other name.

3. Suggest moving the following code inside vmcore.strip_extra_pages since inside that function you're logging other things and you've got the sizes there.
~~~
+            if vmcore.strip_extra_pages():
+                dur = int(time.time() - start)
+                newsize = vmcore_path.stat().st_size
+                log_info("Stripped size: %s" % human_readable_size(newsize))
+                log_info("Makedumpfile took %d seconds and saved %s"
                      % (dur, human_readable_size(oldsize - newsize)))
~~~

4. Another thought I had was is there any point in running makedumpfile if the vmcore is smaller than a certain size, but I'm not suggesting you do it.  I think not but I don't have a good solution for that and don't really think yet another config variable is a good idea.  Maybe a hardcoded minimum but maybe that's not worthwhile either.  Alternatively, it might be worthwhile to have a button to avoid makedumpfile on a vmcore but that's a separate discussion / patch.

Let me do some testing and see if there's anything else I see and let you know.

DaveWysochanskiRH:
If I'm understanding "SpaceSavePercent" correctly, it means "if we save at least this much space, we want to keep the stripped vmcore, otherwise we don't".  If so, then the logic is reversed, and you want < here:
~~~
         elif ((oldsize - newvmcore.stat().st_size) / int(oldsize) * 100 > CONFIG["SpaceSavePercent"]):
~~~

DaveWysochanskiRH:
You might want to improve this message to print the value of SpaceSavePercent (or whatever the final name is)
~~~
[2022-01-19 06:59:38] [W] makedumpfile did not strip enough pages. Not saving new vmcore.
~~~

Maybe something like 
~~~
[2022-01-19 06:59:38] [W] makedumpfile did not strip enough pages (needed 10%). Not saving new vmcore.
~~~

DaveWysochanskiRH:
@audrabaker you can also add this to the bottom near the signature:
Resolves: https://github.com/abrt/retrace-server/issues/445

DaveWysochanskiRH (approved):
This looks good to me.

DaveWysochanskiRH:
@mgrabovsky what do you think?

src/retrace/retrace.py
```diff
@@ -2122,8 +2124,25 @@ def strip_extra_pages(self) -> None:
             log_warn("makedumpfile exited with %d" % child.returncode)
             if newvmcore.is_file():
                 newvmcore.unlink()
+            return 
+
+        #Calculate the % of storage space saved with the new makedumpfile core created. 
+        elif ((oldsize - newvmcore.stat().st_size) / int(oldsize) * 100 < CONFIG["VmcoreDumpSavePercent"]):
```
mgrabovsky, line 2130: The outer parentheses can be removed. Also, change the `elif` to `if`, please, as the previous `if` ends with a `return`. Also, is `int()` necessary in here?

src/retrace/retrace.py
```diff
@@ -2122,8 +2124,25 @@ def strip_extra_pages(self) -> None:
             log_warn("makedumpfile exited with %d" % child.returncode)
             if newvmcore.is_file():
                 newvmcore.unlink()
+            return 
+
+        #Calculate the % of storage space saved with the new makedumpfile core created.
```
mgrabovsky, line 2129: One space after the pound sign please.

src/retrace/retrace.py
```diff
@@ -2122,8 +2124,25 @@ def strip_extra_pages(self) -> None:
...
+                    is: %s" % format_number)
         else:
+            # makedumpfile did strip enough pages so save the new core
             newvmcore.rename(self._vmcore_path)
+            dur = int(time.time() - start)
+            newsize = self.get_path().stat().st_size
+            log_info("Stripped size: %s" % human_readable_size(newsize))
+            log_info("Makedumpfile took %d seconds and saved %s"
+                     % (dur, human_readable_size(oldsize - newsize)))
+        return
```
mgrabovsky, line 2144: The return is unnecessary here.

src/retrace/retrace.py
```diff
@@ -2122,8 +2124,25 @@ def strip_extra_pages(self) -> None:
...
                 newvmcore.unlink()
+            return 
+
+        #Calculate the % of storage space saved with the new makedumpfile core created. 
+        elif ((oldsize - newvmcore.stat().st_size) / int(oldsize) * 100 < CONFIG["VmcoreDumpSavePercent"]):
+            log_warn("makedumpfile did not strip enough pages (needed to save %{}). 
+                    Not saving new vmcore.".format(CONFIG["VmcoreDumpSavePercent"]))
+            format_number = "{:.2f}".format((int(oldsize) - int(newvmcore.stat().st_size)) / int(oldsize) * 100)
+            log_info("Percentage of space that would have been saved if we kept the stripped vmcore 
+                    is: %s" % format_number)
```
mgrabovsky, line 2135: I think this can be simplified slightly:
```suggestion
            saved_pct = (oldsize - newvmcore.stat().st_size) / oldsize * 100
            log_info(f"Keeping the stripped vmcore would have saved {saved_pct:.2f}% of space")
```

src/retrace/retrace.py
```diff
@@ -2122,8 +2124,25 @@ def strip_extra_pages(self) -> None:
             log_warn("makedumpfile exited with %d" % child.returncode)
             if newvmcore.is_file():
                 newvmcore.unlink()
+            return 
+
+        #Calculate the % of storage space saved with the new makedumpfile core created. 
+        elif ((oldsize - newvmcore.stat().st_size) / int(oldsize) * 100 < CONFIG["VmcoreDumpSavePercent"]):
+            log_warn("makedumpfile did not strip enough pages (needed to save %{}). 
+                    Not saving new vmcore.".format(CONFIG["VmcoreDumpSavePercent"]))
```
mgrabovsky, line 2132: String literals cannot be split over lines like this, but you can concatenate them simply by juxtaposition:
```suggestion
            log_warn("makedumpfile did not strip enough pages (needed to save "
                     "at least %{}). Not saving new vmcore.".format(CONFIG["VmcoreDumpSavePercent"]))
```

mgrabovsky (changes requested):
Looks good overall. Just a couple of suggestions.

src/retrace/retrace.py
audrabaker: I found that I could not remove the outer parentheses here unless I wanted to move it to one line. As the rest of the code is all aligned I went ahead and adjusted this so it is on two lines and kept the outer parentheses. Hope that is ok and happy to change it to one line if you want.

DaveWysochanskiRH:
@audrabaker looks like you need to squash the two commits into one, and re-push a single commit:
~~~
$ git log --oneline audrabaker/master | head
f4111cf09c5c Add logic to determine if we save a stripped vmcore
ed0289f90ecd Add logic to determine if we save a stripped vmcore
b564b552e821 github: CD to build directory for checks
~~~

mgrabovsky:
There's an unused variable `oldsize` left on line 793 of `src/retrace/retrace_worker.py` after the refactoring. Once that's removed and squashed, I think we're good to merge.

mgrabovsky (approved):
Looks good to me. Thanks for the patch!

PR #454, opened 2022-01-27
retrace-server-interact: Fix traceback for non-existent gid

DaveWysochanskiRH:
Its possible for a user to be in a group that does not have a name.
In this case retrace-server-interact will Traceback something like
the following
   File "/usr/bin/retrace-server-interact", line 31, in <module>
     groups = [grp.getgrgid(g).gr_name for g in os.getgroups()]
   File "/usr/bin/retrace-server-interact", line 31, in <listcomp>
     groups = [grp.getgrgid(g).gr_name for g in os.getgroups()]
 KeyError: 'getgrgid(): gid not found: 12345'

Fix this by ensuring we don't assume the group database will
have an entry for each gid, in which case getgrgid() will raise
KeyError.  If this happens, catch the exception and just skip
over the gid.

### DIFF
--- a/src/retrace-server-interact
+++ b/src/retrace-server-interact
@@ -28,7 +28,13 @@ def print_cmdline(cmd_line: List[str]):
     sys.stderr.write("$ %s\n\n" % list2cmdline(cmd_line))
 
 if __name__ == "__main__":
-    groups = [grp.getgrgid(g).gr_name for g in os.getgroups()]
+    groups = []
+    for g in os.getgroups():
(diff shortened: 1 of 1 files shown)


DaveWysochanskiRH:
FWW, this fix was verified by user hitting the problem in production.  I think until now this unusual corner case wasn't hit but is a real problem preventing use of retrace-server-interact for that user.

mgrabovsky (approved):
Looks good to me.

Issue #446, opened 2021-11-08
vmcore tasks permanently stuck in 'available' status on manager page for unknown reason


DaveWysochanskiRH:
I've noticed we occasionally get vmcore tasks stuck indefinitely in 'Available'.  I've not been able root cause, but suspected this may be due to improper submission of NULL vmcore paths that would generate tasks that would immediately fail.  I submitted PR https://github.com/abrt/retrace-server/pull/438 to reject these tasks and fail via a HTTP error code, which fixed the tasks that fail in that manner.

Unfortunately that PR didn't fully resolve the problem of "Available" tasks and we are still getting some of these tasks.  One other "early failure" mode of tasks I've seen is tasks failing due to read path permissions issues on the vmcore - retrace-server's httpd or worker tasks do not have access so the task immediately fails with 'no such file or directory'.

I wonder if we could at least fixup the path problem by checking the path inside the httpd process and failing early like that PR 438.  However, I think that might be more risky and end up failing tasks unnecessarily, so needs a bit more thought.

A first step in root causing why the tasks get stuck in 'available' might be some additional debugging as we don't have much info on why these get stuck there.

DaveWysochanskiRH:
This https://github.com/abrt/retrace-server/pull/448 may help since it implements the early permissions check.
In our retrace-server environment, I counted about 25% of failed tasks was due to improper permissions.

mgrabovsky:
@DaveWysochanskiRH Did #448 solve this problem? Can we close this issue now?

DaveWysochanskiRH:
I'm not 100% sure it's solved because it may take some time to show up again and it's been 3 weeks.  I'll close for now and re-open if needed.

DaveWysochanskiRH:
Unfortunately it looks like this is still a problem as I've just got another task stuck in 'available'.  I'll see if I can take a look at this in the next several weeks and come up with a strategy at least on debug, if not a possible patch based on code inspection.

DaveWysochanskiRH:
Ok, I think I see what is happening at least partially.  We had a couple of these 'stuck' tasks, so I went and looked at one of them.  It looks like the tasks get down to the following line(s) where it returns the 'start' URL to the request:
~~~
486     elif filename and filename == "__custom__":
487         POST = request.POST
...
546         task.add_remote(POST["custom_url"])
547         task.set_managed(True)
548         task.set_url("%s/%d" % (match.group(1), task.get_taskid()))
549 
[...]

PR #461, opened 2022-04-02
manager: Prevent double submit of vmcore or coredump tasks

DaveWysochanskiRH:
The POST actions on the manager page to submit a custom vmcore
or coredump should be protected from double clicks which can
cause tasks to be created but never started and end up stuck
in "available" state.  Fix this by disabling the button on
submission.

Fixes: https://github.com/abrt/retrace-server/issues/446


mgrabovsky (approved):
Looks good. Thanks for the PR.

Issue #462, opened 2022-04-03
manager page still displays "List FTP files" at top even when CONFIG["UseFTPTasks"] == 0


DaveWysochanskiRH:
There appears to be no way to disable the "List FTP files" link at the top of the manager page.  This may lead to errors if the link is followed and for some reason the FTP server is no longer available.

PR #463, opened 2022-04-03
manager: Make FTP file box configurable with UseFTPTasks

DaveWysochanskiRH:
If CONFIG["UseFTPTasks"] is set to 0 then we should not waste
space on the manager page with the "List FTP files" link.

Fixes: https://github.com/abrt/retrace-server/issues/462


src/manager.wsgi
```diff
@@ -948,6 +948,13 @@ def application(environ, start_response):
 
     custom_url = "%s/__custom__" % match.group(1)
 
+    ftp_box = ""
+    if CONFIG["UseFTPTasks"]:
+        with open("/usr/share/retrace-server/manager_ftp_box.xhtml") as file:
+            # Read 1 MiB
+            ftp_box = file.read(1 << 20)
```
mgrabovsky, line 955: I think I'd prefer the HTML fragment to stay in this file in some variable/constant. It seems a bit excessive to create a special file for such a small chunk.

src/manager.xhtml
```diff
@@ -214,9 +214,7 @@
     </head>
     <body>
         <script src="//ajax.googleapis.com/ajax/libs/jquery/1.9.1/jquery.min.js"></script>
-        <div id="ftp-wrapper">
-            <h2><a href="ftp">List FTP files (create vmcore task from FTP)</a></h2>
-        </div>
+	{ftp_box}
```
mgrabovsky, line 217: This should be indented wit spaces.

src/manager.wsgi
DaveWysochanskiRH: Great feedback - fixed in 3a31b86.

src/manager.xhtml
DaveWysochanskiRH: Fixed in 3a31b86.

PR #476, opened 2022-10-06
retrace-server-task: Add exception handling for SSLError

DaveWysochanskiRH:
It's not uncommon in production to have a certificate expire, and when that happens we use the FAILBACK_SERVER, but this is unlikely to be what the user wants.

This patch catches the error and gives the user possible next steps to avoid the error, similar to earlier commit cf85a18c1b29.

### DIFF
--- a/src/retrace-server-task
+++ b/src/retrace-server-task
@@ -707,6 +707,12 @@ if __name__ == "__main__":
             sys.exit(1)
         if req.status_code != 200:
             raise requests.exceptions.ConnectionError
+    except requests.exceptions.SSLError as ex:
+        if "debug" in ARGS:
+            LOGGER.error("SSLError: '%s'", str(ex))
+        LOGGER.error("SSL error connecting to {}: try adding '-v' to see details"
+                     " or '--no-verify' for an expired certificate.".format(ARGS["server"]))
+        sys.exit(1)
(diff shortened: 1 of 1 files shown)


DaveWysochanskiRH:
I looked into catching ssl.CertificateError but I could not make that work at all, so opted for the more generic SSLError.

I was also looking at ways to remove the FALLBACK_SERVER completely - maybe we change this to a config variable "DEFAULT_SERVER" or something like that, and then pick a default.  I don't mind having that set on all the production systems.  For our production systems we don't need the FALLBACK_SERVER and as I recall this was a tradeoff between two users who each wanted a different default server if none was given.  I think it's a different patch but was wondering what others thoughts were on it, if any.

src/retrace-server-task
```diff
@@ -17,6 +17,7 @@ from typing import Any, Dict, Optional, Tuple
 import requests
 from requests.packages.urllib3.exceptions import InsecureRequestWarning
 from requests_gssapi import HTTPSPNEGOAuth, DISABLED
+from requests.exceptions import SSLError
```
mgrabovsky, line 20: This import is unnecessary since you refer to the exception by the fully qualified name down below.

DaveWysochanskiRH: Removed

PR #477, opened 2022-10-17
stats: Fix Releases counting by using wildcard at end and add el9

DaveWysochanskiRH:
The stats database has a 'version' field which records the version of the package.  The value of this field though may not end in the any of the strings listed in the plugins though.  For the kernel, the 'arch' value is included at the end of the value.  Also, for any z-stream package versions include the minor (Y) version after the major (X) version (for example, "el8_2" for a 8.2.0.z package). Fix these issues by using a wildcard at the end of the database query string.

In addition, add 'el9' to the 'rhel' and 'centos' plugins.

### DIFF
--- a/src/plugins/centos.py
+++ b/src/plugins/centos.py
@@ -15,5 +15,6 @@ versionlist = [
     "el6",
     "el7",
     "el8",
+    "el9",
 ]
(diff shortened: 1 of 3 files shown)


mgrabovsky (approved):
Looks OK.

PR #488, opened 2023-03-31
retrace: Remove low value 'not an archive' message for vmcores

DaveWysochanskiRH:
The message to log 'not an archive' is of little to no value. In our production system we see this logged for every vmcore so it just takes up log space.  Also there is no message logged for coredumps so remove this message for vmcores.


mgrabovsky (approved):
Makes sense.

